PR #1, opened 2018-09-14
Avoid first lines of paragraphs to be output as code

kabir:


kabir:
@bshannon I have fixed the commit.

PR #12, opened 2018-11-13
Remove RI folder if it already exists

ggam:
When running all tests locally in sequence, the unzipped GlassFish was preserved, preventing a clean installation. Also, the shell required manual interaction to confirm you want to overwrite the files.


arjantijms:
Might also want to add an option here to not download latest-glassfish all the time if it's already there?

PR #13, opened 2018-11-13
Fixed TS_HOME env variable

ggam:
At least on my local machine the correct path seems to be `/install/j2ee` inside the cloned repository. Also it shouldn't replace the variable if it was already defined.

### DIFF
--- a/docker/run_jakartaeetck.sh
+++ b/docker/run_jakartaeetck.sh
@@ -34,7 +34,11 @@ fi
 if [ -z "${CTS_HOME}" ]; then
   export CTS_HOME="${WORKSPACE}"
 fi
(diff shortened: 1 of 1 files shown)


alwin-joseph:
1. TS_HOME is now set as "export TS_HOME=${CTS_HOME}/jakartaeetck" (renamed javaeetck to jakartaeetck)
2. TS_HOME is the home directory where the TCK is unzipped, it is not the install/jakartaee directory(install/javaee renamed to install/jakartaee). All the variables that rely on TS_HOME assumes that TS_HOME is the ${CTS_HOME}/jakartaeetck/ directory, for the tests run from docker/run_jakartaeetck.sh

Can you please cross check this and close this.

PR #42, opened 2019-02-07
Fix #41 correct HTTP header checks that are case sensitive for header names

### LINKED ISSUE #41: Servlet TCK serverpushInitTest uses case sensitive test for HTTP headers
HTTP header names are case insensitive. This test (full name below) performs case sensitive tests which will fail if the container uses, for example, all lower case HTTP header names.

`com/sun/ts/tests/servlet/spec/serverpush/Client.java#serverPushInitTest`

PR to follow shortly.

markt-asf:


markt-asf:
I can remove the whitespace changes if that helps (and the restore the unused imports).

The logic is simpler if you change the header case in the response body. There is no String.containsIgnoreCase(...) and if you start forcing everything to lower case on the client to make the header name comparisons case insensitive you risk corrupting values like the URI which are case sensitive. All of that is solvable on the client but would require rather more code to parse out the header names.

bshannon:
(It's generally better to do code cleanup like white space changes in a separate commit/PR.)

You could use response.toLowerCase(Locale.ENGLISH).contains("range").

Or add

```
private static boolean containsIgnoreCase(String s1, String s2) {
    return s1.toLowerCase(Locale.ENGLISH).contains(s2.toLowerCase(Locale.ENGLISH));
}
```

markt-asf:
Agreed on the clean-up. It wasn't intentional. My IDE has some auto-save actions configured by default and I didn't notice the impact until it was too late. I've turned them off for this project now and pushed an updated PR.

Converting the whole response to lower case is OK for the simple header checks but I don't like it for the referrer because the URI is case sensitive. Granted a case insensitive check on the URI is probably going to be OK but I don't like the idea of introducing one bug to fix another since it would allow a pass if the container illegally changed the URI case in the referrer header. Unlikely but you never know...

bshannon:
Uncle!

It still seems like a kludge to "fix" this on the server side.  We could do the client side
matching with regular expressions, but I think we're past the point of diminishing returns.

PR #55, opened 2019-06-03
changes for running interop against GF 5.0.1 as RI and latest eclipse GF as VI

ankitoracle:


alwin-joseph:
Ankit can you merge this on EE4J_8 branch.
Changes look good to me, we would need to make the ci job changes to provide GF 5.0.1 URL, after this merge.

gurunrao:
@ankitoracle - please name glassfish-5.0.zip to older-glassfish.zip or reference-glassfish.zip - so that  change can be generic in nature and not tied to 5.0/5.1

PR #61, opened 2019-06-18
User guide build updates

bshannon:


bshannon:
1. staging contains the html version; generated-docs contains the pdf file (the other files in that directory can be ignored).  The book directory is a temporary directory used during the generation of the pdf version.
2. That's the point at which it generates the pdf file.  It can take close to a minute.  It works for me.
3. Yes, there are some formatting errors that will need to be fixed.  I've mostly left these to the project teams to fix later.
4. Ditto.
5. Yes, I see that now.  I've pushed a fix.

Yes, it would be nice if the title contained the TCK name, but that requires more configuration work that I left until later.

edbratt:
"Java EE Licensees" is one of those substitutions that need to be made. 
Alwin, if that's part of your concern, I will try to start working on those.

PR #63, opened 2019-06-26
Changes to create new Bundle with Eclipse License & /docs to have generated userguide.

alwin-joseph:


alwin-joseph:
@bshannon The LICENSE_EFTL.md file is now only added to the repo root directory. Each TCK has  LICENSE.md file in install/<tck>/ directory which is the file that is copied to the respective standalone TCK bundle. Please let me know if we need to copy the LICENSE_EFTL.md file to each of the install/<tck> directory too.

bshannon:
> @bshannon The LICENSE_EFTL.md file is now only added to the repo root directory. Each TCK has LICENSE.md file in install/tck/ directory which is the file that is copied to the respective standalone TCK bundle. Please let me know if we need to copy the LICENSE_EFTL.md file to each of the install/tck/ directory too.

If there was a possibility that the license would be different for each TCK, then having a copy of the license in each install/tck directory would make sense.  But that's not the case here.  We don't need multiple copies of either the original EPL LICENSE.md file, nor of the LICENSE_EFTL.md file.

alwin-joseph:
> As far as I can tell without doing a build myself, this looks like it will do the right thing.
> 
> It's a bit confusing using "Eclipse" as the name of the other license since _both_ licenses are Eclipse licenses. It would be better to use "EFTL" as the property name to select use of the EFTL license.
> 
Done.
> There's a bunch of file renames here that I assume are really directory renames that git is confused about.
> 
yes. 
> Is there really no way to refactor the release/tools/*.xml files so that the same changes don't need to be duplicated to all of them? Couldn't there be one top level xml file that sets parameters and then includes a common rules xml file?
That will involve a lot of initial changes and refactoring. Skipping that for now. Bundle creation of each tck happens at the corresponding release/tools/*.xml file.

PR #81, opened 2019-07-18
fix to remove autogenerated .oracle_jre_usage directory in the TCK bundles

alwin-joseph:


alwin-joseph:
The .oracle_jre_usage directory is automatically recreated by the system even after deleting it with this change. Revoking this PR as this change is not sufficient for the fix.

bshannon:
You probably just need to exclude it when creating the zip file.

PR #82, opened 2019-07-19
fix to remove auto-generated ?/.oracle_jre_usage dirs

alwin-joseph:


alwin-joseph:
> Isn't there a way to update the ant zip rule so that it excludes this directory to begin with, rather than deleting it after the fact?

The creation of this '?' directory seem unpredictable. I do not see it in my local where I used jdk8_121. As i understand its an issue specific to the jdk version and occurs when run in the current CI jobs. Removing it after the bundle generation seemed safer without affecting the current ant tasks. But Ill try to exclude it from the ant task that creates the  zip.

PR #87, opened 2019-07-19
Add some missing rules and definitions to the template.

bshannon:
Apply template.
Update jaxws and jsp.


arjantijms:
Good work!

PR #90, opened 2019-07-24
change the standalone TCK bundle names as per Spec Committee

alwin-joseph:

### DIFF
--- a/docker/cajtck.sh
+++ b/docker/cajtck.sh
@@ -23,11 +23,11 @@ echo "ANT_OPTS in cajtck.sh $ANT_OPTS"
 cd $TCK_HOME
 
 if ls ${WORKSPACE}/standalone-bundles/*annotations-tck*.zip 1> /dev/null 2>&1; then
-  echo "Using stashed bundle created during the build phase"
+  echo "Using stashed bundle for annotations-tck created during the build phase"
   unzip ${WORKSPACE}/standalone-bundles/*annotations-tck*.zip -d ${TCK_HOME}
   TCK_NAME=annotations-tck
 elif ls ${WORKSPACE}/standalone-bundles/*cajtck*.zip 1> /dev/null 2>&1; then
-  echo "Using stashed bundle created during the build phase"
+  echo "Using stashed bundle for cajtck created during the build phase"
   unzip ${WORKSPACE}/standalone-bundles/*cajtck*.zip -d ${TCK_HOME}
   TCK_NAME=cajtck
 else
--- a/docker/cajtck.sh
+++ b/docker/cajtck.sh
@@ -22,9 +22,14 @@ echo "ANT_OPTS in cajtck.sh $ANT_OPTS"
 
 cd $TCK_HOME
 
-if ls ${WORKSPACE}/standalone-bundles/*cajtck*.zip 1> /dev/null 2>&1; then
+if ls ${WORKSPACE}/standalone-bundles/*annotations-tck*.zip 1> /dev/null 2>&1; then
+  echo "Using stashed bundle created during the build phase"
+  unzip ${WORKSPACE}/standalone-bundles/*annotations-tck*.zip -d ${TCK_HOME}
+  TCK_NAME=annotations-tck
+elif ls ${WORKSPACE}/standalone-bundles/*cajtck*.zip 1> /dev/null 2>&1; then
   echo "Using stashed bundle created during the build phase"
   unzip ${WORKSPACE}/standalone-bundles/*cajtck*.zip -d ${TCK_HOME}
+  TCK_NAME=cajtck
 else
   echo "[ERROR] TCK bundle not found"
   exit 1
@@ -39,7 +44,8 @@ fi
 wget --progress=bar:force --no-cache $GF_BUNDLE_URL -O latest-glassfish.zip
 unzip ${TCK_HOME}/latest-glassfish.zip -d ${TCK_HOME}
 
-TS_HOME=$TCK_HOME/cajtck
+TS_HOME=$TCK_HOME/$TCK_NAME
+
 echo "TS_HOME $TS_HOME"
 
 chmod -R 777 $TS_HOME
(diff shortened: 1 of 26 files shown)


alwin-joseph:
The changed names of the TCKs are as below in reference to https://docs.google.com/spreadsheets/d/18SraPxRBCOyaS6w-UV6TR-UA1bWy1--sV0ky6msAjWY/edit#gid=0

caj - annotations
connector - connectors
el - expression-language
jacc - authorization
jaspic - authentication
jaxrpc - xml-rpc
jaxrs - restful-ws
jaxr - xml-registries
jaxws - xml-ws
jms - messaging
jpa - persistence
jsf - faces
jsp - pages
jstl - tags
jta - transactions
saaj - soap
securityapi - security

arjantijms:
For the specs I'm involved with all seems fine. Thx @alwin-joseph 

- ✔︎ jacc - authorization
- ✔︎ jaspic - authentication
- ✔︎ securityapi - security
- ✔︎ el - expression-language
- ✔︎ jsf - faces

docker/cajtck.sh
```diff
@@ -48,24 +54,23 @@ cd $TS_HOME/bin
 sed -i "s#^build.level=.*#build.level=2#g" ts.jte
 sed -i "s#^endorsed.dirs=.*#endorsed.dirs=$TCK_HOME/glassfish5/glassfish/modules/endorsed#g" ts.jte
 sed -i "s#^local.classes=.*#local.classes=$TCK_HOME/glassfish5/glassfish/modules/endorsed/jakarta.annotation-api.jar#g" ts.jte
-sed -i "s#^report.dir=.*#report.dir=$TCK_HOME/cajtckreport/cajtck#g" ts.jte
-sed -i "s#^work.dir=.*#work.dir=$TCK_HOME/cajtckwork/cajtck#g" ts.jte
+sed -i "s#^report.dir=.*#report.dir=$TCK_HOME/${TCK_NAME}report/${TCK_NAME}#g" ts.jte
+sed -i "s#^work.dir=.*#work.dir=$TCK_HOME/${TCK_NAME}work/${TCK_NAME}#g" ts.jte
```
bshannon: A nit, but I would probably add a dash in front of "report" and "work".

bshannon: Actually, why does TCK_NAME need to appear before "report" or "work" at all
since it occurs in the directory name after "report" and "work"?  For that matter,
why does it need to appear in the directory name at all?  Maybe it's needed to
allow multiple TCKs to run in the same directory?

For Jakarta Mail, the directories are JTreport and JTwork, with no TCK_NAME
in the directory name at all.

docker/cajtck.sh
```diff
@@ -48,24 +54,23 @@ cd $TS_HOME/bin
...
+mkdir -p $TCK_HOME/${TCK_NAME}work/${TCK_NAME}
 
 cd $TCK_HOME/glassfish5/bin
 ./asadmin start-domain
 
 cd $TS_HOME/src/com/sun/ts/tests/signaturetest/caj
 ant runclient
 
-TCK_NAME=cajtck
 JT_REPORT_DIR=$TCK_HOME/${TCK_NAME}report
```
bshannon: Another nit...  JT_REPORT_DIR could be set earlier, and could be used in more places in the script.
It might also be helpful to define a JT_WORK_DIR similarly.

release/tools/build.xml
```diff
@@ -171,8 +189,28 @@
...
+        <!--property name="zip.name" value="${deliverable.tck.dir.token}-${deliverable.type}"/-->
+	<condition property="zip.name" value="${deliverable.name.token}${deliverable.type}">
+            <contains string="${deliverable.name}" substring="jakartaee" casesensitive="false"/>
+        </condition>
+        <condition property="zip.name" value="${deliverable.tck.dir.token}-${deliverable.type}">
+            <not>
+                <contains string="${deliverable.name}" substring="jakartaee" casesensitive="false"/>
+            </not>
+        </condition>
+
```
bshannon: I can't quite tell why jakartaee needs to be handled differently here.
Maybe a comment with examples would make it clear?

alwin-joseph:
> cajtck.sh: TCK_NAME should be annotations-tck
Same is used.
> General comment, I'm not sure when hyphens are used or not used. My suggestion is to be consistent. Either always use them, or never use them. My preference it to use them for readability, but you may decide. Ref: connectortck.sh line 28 vs line 32.
connector tck bundle will be named "connector-tck_1.7.0.zip" , hyphen is used.
> connectortck.sh -- Not sure why you need the extra directory at lines 59 & 60, but I guess that's what was there before.
The two directories are created - as report & work directories. The same is added to ts.jte.
> docker/jacctck.sh: new lines 28 vs 32. I wonder why wouldn't the build-phase create the correct name? That change will come later? I would recommend that the report/log text (e.g. docker/jaxrstck.sh) at lines 26 and 30 be different. Otherwise the log output won't distinguish which if clause is being used.
build-phase will create the correct bundles. The additional "if" is to handle old bundles to be run from CI.
ill change the log output to distinguish.
> in release/tools/build.xml, please confirm that property "deliverable.tck.dir" is not necessary for TCKs that are not changing (i.e. jsonb, jsonp, ...).
Yes, "deliverable.tck.dir" is not necessary for TCKs that are not changing

docker/cajtck.sh
alwin-joseph: > Actually, why does TCK_NAME need to appear before "report" or "work" at all
> since it occurs in the directory name after "report" and "work"? For that matter,
> why does it need to appear in the directory name at all? Maybe it's needed to
> allow multiple TCKs to run in the same directory?
${TCK_NAME}report is required when the TCKs share the same workspace. ${TCK_NAME}report/${TCK_NAME} may not be required, but let me keep it for now.
> 
> For Jakarta Mail, the directories are JTreport and JTwork, with no TCK_NAME
> in the directory name at all.

docker/cajtck.sh
alwin-joseph: > Another nit... JT_REPORT_DIR could be set earlier, and could be used in more places in the script.
> It might also be helpful to define a JT_WORK_DIR similarly.
Will make that change.

release/tools/build.xml
alwin-joseph: We use these same ant tasks to create jakartaeetck.zip (& jakartaee-smoke.zip) bundles. I did not want to apply the TCK bundle name changes to those bundles now.  This change is to continue using the old names for those bundles.

docker/cajtck.sh
alwin-joseph: > > Another nit... JT_REPORT_DIR could be set earlier, and could be used in more places in the script.
> > It might also be helpful to define a JT_WORK_DIR similarly.


Will make this improvement later.

PR #127, opened 2019-10-03
Revert "Reverting commits that changes the test behaviour compared to JavaEE8"

alwin-joseph:
Reverts eclipse-ee4j/jakartaee-tck#99

PR #129, opened 2019-10-07
Revert "Revert "Reverting commits that changes the test behaviour compared to JavaEE8""

alwin-joseph:
Reverts eclipse-ee4j/jakartaee-tck#127

This is to pause the contributions as instructed by Corp Arch until said otherwise.

Just to avoid confusion with the reverts, after this PR merge the improvements to the tests will no longer be in place. We will create another PR to include them.

### DIFF
--- a/src/com/sun/ts/tests/servlet/api/javax_servlet_http/cookie/TestServlet.java
+++ b/src/com/sun/ts/tests/servlet/api/javax_servlet_http/cookie/TestServlet.java
@@ -72,8 +72,6 @@ import java.util.TimeZone;
 
 public class TestServlet extends HttpTCKServlet {
 
-  public static String CUSTOM_HEADER_DATE_FORMAT = "yyyy-MM-dd HH:mm";
-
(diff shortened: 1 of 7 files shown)


markt-asf:
Who or what is "Corp Arch" and where do they / does it sit in the Eclipse project governance structure?

alwin-joseph:
@markt-asf  please ignore my previous comment.

markt-asf:
OK, consider it ignored.

Can you expand on the "needs more review"? These changes have been applied, reverted, re-applied and now re-reverted. I can understand the reasoning for the first revert (keep Jakarta EE 8 aligned with Java EE 8) - even while I disagree with reasoning - because the issue was discussed on the dev@ list.

My understanding from the discussion on #127 was that these changes were OK for master on the basis that any further Jakarta EE 8 releases would be on a branch based off the 8.0.1 tag.

I was, therefore, surprised to see this revert appear out of the blue as it were. Can you shed some light on what is going on?

PR #135, opened 2019-11-10
[#554] Added test for Response.created() with relative URI

ronsigal:
Signed-off-by: Ron Sigal <rsigal@redhat.com>

### DIFF
--- a/src/com/sun/ts/tests/jaxrs/ee/rs/core/response/JAXRSClient.java
+++ b/src/com/sun/ts/tests/jaxrs/ee/rs/core/response/JAXRSClient.java
@@ -1552,6 +1552,29 @@ public class JAXRSClient extends JaxrsCommonClient {
     }
   }
 
+    /*
+   * @testName: responseCreatedRelativeURITest
+   *
+   * @assertion_ids: JAXRS:JAVADOC:121;
+   *
+   * @test_Strategy: The resource calls Response.created() to set the Location header with a
+   * relative URI. The relative URI should be converted into an absolute URI by resolving it
+   * relative to the base URI.
+   */
+  public void responseCreatedRelativeURITest()
+      throws Fault {
+    String resourceUrl = getAbsoluteUrl();
+    String expected = resourceUrl.substring(0, resourceUrl.length() - "resource".length()) + "created";
+    Response response = invokeGet("created");
+    try {
+      assertFault(expected.equals(response.getHeaderString("location")),
+        "#response.getHeaderString(\"location\") [" +
+        response.getHeaderString("location") + "] differs from ", expected);
+    } finally {
+      response.close();
+    }
+  }
+
   // ////////////////////////////////////////////////////////////////////
   protected <T> GenericType<T> generic(Class<T> clazz) {
     return new GenericType<T>(clazz);
--- a/src/com/sun/ts/tests/jaxrs/ee/rs/core/response/ResponseTest.java
+++ b/src/com/sun/ts/tests/jaxrs/ee/rs/core/response/ResponseTest.java
@@ -277,6 +277,17 @@ public class ResponseTest {
     return response;
   }
 
+  @Path("created")
+  @GET
+  public Response setLocationHeader() {
+    try {
+      Response response = Response.created(new URI("created")).status(200).build();
+      return response;
+    } catch (URISyntaxException e) {
+       throw new RuntimeException(e);
+    }
+  }
+
   // ////////////////////////////////////////////////////////////////////
   protected static Link createLink(String path, String rel) {
     return Link.fromUri(createUri(path)).rel(rel).build();


ronsigal:
This pull request is in support of https://github.com/eclipse-ee4j/jaxrs-api/pull/810.

bshannon:
@jansupol , please review this PR.

src/com/sun/ts/tests/jaxrs/ee/rs/core/response/JAXRSClient.java
```diff
@@ -1552,6 +1552,29 @@ public void readEntityGenericTypeAnnotationThrowsIllegalStateExceptionTest()
...
+   * relative URI. The relative URI should be converted into an absolute URI by resolving it
+   * relative to the base URI.
+   */
+  public void responseCreatedRelativeURITest()
+      throws Fault {
+    String resourceUrl = getAbsoluteUrl();
+    String expected = resourceUrl.substring(0, resourceUrl.length() - "resource".length()) + "created";
+    Response response = invokeGet("created");
+    try {
+      assertFault(expected.equals(response.getHeaderString("location")),
```
jansupol: RFC 7231 allows for using both absolute and relative URLs in the HTTP location header. I assume that in the case of the relative URL this exact matching would not work.

I assume that a test for `response.getHeaderString("location").contains("created")` would work here.

ronsigal:
Hi @jansupol,

I don't really understand your question. 

1. The javadoc for Response.created() says
```
     * @param location the URI of the new resource. If a relative URI is
     *                 supplied it will be converted into an absolute URI by resolving it
     *                 relative to the request URI (see {@link UriInfo#getRequestUri}).
```
so there shouldn't be a location header with a relative URL coming back.

2. Actually, in this test, the resource method IS passing a relative URL into Response.created() - not sure if that's part of your question.

3. I would also add that com.sun.ts.tests.jaxrs.api.rs.core.responseclient.JAXRSClient.createdTest() calls Response.created() with a variety of absolute and relative URLs. 

Note that the latter test is more of a unit test in that it doesn't run in the context of a server. I suppose one could argue that there might be a test like responseCreatedRelativeURITest() but which passes an absolute URL into Response.created() on a server. But that's not really relevant to the issue I'm addressing.

-Ron

jansupol:
@ronsigal Sorry, I missed that part of the Javadoc.

scottmarlow:
@alwin-joseph asked on jakartaee-tck-dev@eclipse.org if this pull request (that adds a new test) is targeting Jakarta EE 8.  I think that the code change was made for that intention, but targeting Jakarta EE 9 would be better, so that we don't risk breaking existing Jakarta EE 8 compatible implementations with a new test.

Should we create a label for pull requests that we want to merge for EE9?

CC @jansupol

ronsigal:
I would say that the test is safe to add to Jakarta EE 8:

  1. The related change to the javadoc (for Response.created()) was to 
make the javadoc conform to the existing JAX-RS code behavior.

  2. The TCK test is a regression test that just confirms that the 
existing JAX-RS code continues to behave the way it already does.

I'm just saying this to clarify the situation. I don't there's any harm 
in delaying the test until Jakarta EE 9.

-Ron

jansupol:
@ronsigal This is more about consistency. If the test is added to Jakarta EE 8, then it would be precedence allowing new tests in the existing version of any Specification. I do not say it's wrong or not, but the rule should be clear to follow. A compatible implementation may become not compatible with a new test. While this particular test can be of no harm, in general, new tests can be harmful to vendors when added to the existing version of the Specification. The rule should be to allow it or not, but mixing it seems confusing.

jansupol:
@scottmarlow It could make sense to have two branches, one for Jakarta EE 8 patches and one for Jakarta EE 9 (Though the patches should go to both, assumingly their number ought to be low). We have just a single (master) branch, so the label looks like a solution (Maybe more generic NewTest label?).

ronsigal:
Ok, thanks for the clarification, @jansupol.

bshannon:
> @scottmarlow It could make sense to have two branches, one for Jakarta EE 8 patches and one for Jakarta EE 9 (Though the patches should go to both, assumingly their number ought to be low). We have just a single (master) branch, so the label looks like a solution (Maybe more generic NewTest label?).

Having two branches is fine, as long as one of them is "master".

For example, use master for the currently under development version for Jakarta EE 9,
and use a "2.3-PATCH" branch (or whatever the right number is) for patches against
the previous release (largely only updates to the exclude list).  The exclude list updates
shouldn't (generally) be merged into master because the bug should just be fixed in master.
(If the bug is too hard to fix, it might remain excluded.)

scottmarlow:
The CI failure was a "java.nio.channels.ClosedChannelException", doesn't seem related to this change.  Does anyone know how to force the "continuous-integration/jenkins/pr-merge" testing to run again from this pull request?

scottmarlow:
Regarding where to make EE 8 platform TCK patches, we have the 8.0 branch for excluding challenged tests, not that this pr changes any existing tests.

scottmarlow:
Merging this, so we will see test results with GlassFish 5.1 tonight.

PR #144, opened 2020-01-06
Register consumer before the EventSource is open not to loose the message

jansupol:
Signed-off-by: Jan Supol <jan.supol@oracle.com>


jansupol:
The first message gets intermittently lost in between the source open and consumer registration as of now.

PR #150, opened 2020-02-03
Updates for EL 4.0

markt-asf:
The key changes are in the first commit. The second is just a modification I made to the build script that I found useful.

I replaced the jakarta.el.jar locally in glassfish with [jakarta.el-api-4.0.0-RC1](https://repo1.maven.org/maven2/jakarta/el/jakarta.el-api/4.0.0-RC1/jakarta.el-api-4.0.0-RC1.jar) from Maven central to build this.

I ran the resulting TCK bundle with jakarta.el-api-4.0.0-RC1 and the Apache Tomcat implementation. The TCK passes when using both Java 8 and Java 11.

I had to make some changes to the shared TCK infrastructure to get the Java 11 signature tests working.

The signature tests failed with Java 11 unless I replaced lib/sigtest.jar with the [latest snapshot build](https://ci.adoptopenjdk.net/view/all/job/sigtest/)
The error message in the logs was `Unknown option -xjimage.`

I'll be surprised if this PR is perfect first time around. Happy to adjust it as required after review.


bshannon:
Generally we haven't been removing the old signature files, but you might say the
Java EE - > Jakarta EE transition warrants a special case.  Ideally this would be a
policy applied consistently across all the specs.  Maybe this should be brought up
at the Platform project team?

kwsutter:
> Generally we haven't been removing the old signature files, but you might say the
> Java EE - > Jakarta EE transition warrants a special case. Ideally this would be a
> policy applied consistently across all the specs. Maybe this should be brought up
> at the Platform project team?

@bshannon Was there a reason to keep the old signature files?  I looked at the repo and it seems the most components kept the last few iterations of their signature files.  Since there's a new revision of the TCK with each major release, is there a reason to keep the old signature files?  And, as you mentioned, especially with the move from `javax` to `jakarta`, it makes even less sense.

bshannon:
I checked with one of the original CTS engineers and he doesn't know of any particular reason that the old signature files were kept, so I think a policy of removing them and only keeping the current one makes sense.

kwsutter:
Per the [discussion on our JakartaEE TCK dev](https://www.eclipse.org/lists/jakartaee-tck-dev/msg00372.html) mailing list, I added [a FAQ entry](https://github.com/eclipse-ee4j/jakartaee-platform/pull/166) to explain that the old `javax` signature files do not have to be kept around.  I will also go ahead with this merge since @scottmarlow has approved it.

scottmarlow:
FYI, https://ci.eclipse.org/jakartaee-tck/job/jakartaee-tck/job/master/524 is running now (using last nights build of the Platform TCK) which should include this change.

Issue #151, opened 2020-02-05
Different results with JDK8 and JDK11 with date format in german locale


jbescos:
There is one test failing in TCK related to this topic, when it is executed with JDK11. The test is: 'com.sun.ts.tests.jsonb.api.annotation.AnnotationTest.testJsonbDateFormat'

To illustrate the issue:

You can run the next test with JDK8 and JDK11, and you will see that is success with JDK8 but it doesn't work for JDK11:
`
public class TckTest {
    
    private static final String EXPECTED_DATE = "Do, 01 Jan 1970";
    private static final String DATE_PATTERN = "EEE, dd MMM yyyy";
    private static final String LOCALE_DE = "de";
    private static final Date date = new Date(0);
    
    @Test
    public void testDateTimeFormatter() {
        DateTimeFormatter dateFormatter = DateTimeFormatter.ofPattern(DATE_PATTERN, Locale.forLanguageTag(LOCALE_DE));
        String dateFormatted = dateFormatter.format(date.toInstant().atOffset(ZoneOffset.UTC).toLocalDateTime());
        assertEquals(EXPECTED_DATE, dateFormatted);
    }
    
    @Test
    public void testSimpleDateFormat() {
        DateFormat dateFormatter = new SimpleDateFormat(DATE_PATTERN, Locale.forLanguageTag(LOCALE_DE));
        String dateFormatted = dateFormatter.format(date);
        assertEquals(EXPECTED_DATE, dateFormatted);
    }
    
}
`

In JDK11 both fails with the next (note there are dots after Do and Jan):
expected:<Do[, 01 Jan] 1970> but was:<Do[., 01 Jan.] 1970>

You can see another example here:
https://stackoverflow.com/questions/50411064/simpledateformat-with-german-locale-java-8-vs-java-10

I think this should be fixed in tck, to support both (with and without dots).

jbescos:
I have created this PR with my suggestion:
https://github.com/eclipse-ee4j/jakartaee-tck/pull/152/

PR #152, opened 2020-02-05
Different results with JDK8 and JDK11 with date format in german locale

jbescos:
This PR is to fix this issue:
https://github.com/eclipse-ee4j/jakartaee-tck/issues/151

### DIFF
--- a/src/com/sun/ts/tests/jsonb/api/annotation/AnnotationTest.java
+++ b/src/com/sun/ts/tests/jsonb/api/annotation/AnnotationTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017, 2018 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2017, 2020 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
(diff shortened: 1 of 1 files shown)


alwin-joseph:
@jbescos  Can this change help pass the test in both JDK8 & JDK11  or is it specific to JDK11?

jbescos:
@alwin-joseph it will pass with both.

bshannon:
Is this fix only for Jakarta EE 9, or should it be applied to Jakarta EE 8 as well?

jbescos:
@bshannon this fix is not related to jakarta, it is a problem with the locales in German. I think you are asking because of the other PR I have related to jakarta, but this new PR is unrelated.

Find in my example of the issue https://github.com/eclipse-ee4j/jakartaee-tck/issues/151 that the problem is reproducible using SimpleDateFormat and DateTimeFormatter.

This fix make it work from JDK8 to JDK11 (included).

bshannon:
So to be clear, this change is applicable to the Jakarta EE 9 TCK (master branch of the TCK repo)
when using either JDK 8 or JDK 11, right?  If so, then this should be approved and merged.

What's not clear is whether this problem also occurs with the Jakarta EE 8 TCK when run with either
JDK 8 or JDK 11.  If it does apply, then someone needs to file a test challenge against Jakarta
EE 8 so that we can exclude the test from Jakarta EE 8 (we're not fixing tests in Jakarta EE 8).
Although if it does apply, I'm not sure why no one has reported it yet.  Was this a new test that
was added for Jakarta EE 9?

PR #154, opened 2020-02-05
Proposal for aggregating external TCK tests

aguibert:
On the jakartaee-dev mailing list we have been discussing splitting out some specs (such as JSON-B) to house their TCK tests in the same repo where the api/spec live. See [this PR ](https://github.com/eclipse-ee4j/jsonb-api/pull/221)for an example.

This PR sets up the framework for aggregating external TCK tests into a single place so they can be easily run. The current modules I've included are BeanValidation and JSON-B (but JSONB will only work if you build it locally for now). Eventually we could include CDI also.


external-tcks/README.md
```diff
@@ -0,0 +1,27 @@
+# JakartaEE TCK tests from extenral sources
```
arjantijms: Typo: external

aguibert:
hi @bshannon 

> How does the external test suite get run in all the different containers by the JavaTest harness,
like it did before? Does it produce the same output as before?

the external test suite now runs with Arquillian+JUnit instead of JavaTest. The output will be the standard JUnit test result output.

> I thought we had decided that the configuration to run the test suite in some other app server
product would not be included in this repository.

That's what I was trying to ask on the mailing list -- I didn't think it made sense to include all of the various app server configs in this repo either. I'd be happy to provide instructions on how to run these new external test suites instead of app server config profiles.

Overall I'm not proposing anything radically new. I'm just proposing that specs who want to use more modern/familiar test frameworks such as Arquillian+JUnit can do so. The precedent has already been set by the CDI and Bean Validation specs (except they use TestNG instead of JUnit), and I find the CDI/BeanVal way of doing this to be much easier to work with and maintain, so I'd like to do the same with JSON-B (and set an example for other specs too)

bshannon:
> hi @bshannon
> 
> > How does the external test suite get run in all the different containers by the JavaTest harness,
> > like it did before? Does it produce the same output as before?
> 
> the external test suite now runs with Arquillian+JUnit instead of JavaTest. The output will be the standard JUnit test result output.

So in other words, it's not doing the same testing as the previous test, right?

> > I thought we had decided that the configuration to run the test suite in some other app server
> > product would not be included in this repository.
> 
> That's what I was trying to ask on the mailing list -- I didn't think it made sense to include all of the various app server configs in this repo either. I'd be happy to provide instructions on how to run these new external test suites instead of app server config profiles.

(Sorry, I ran across this PR before I saw the mailing lit conversation.)

I think we want instructions on how to configure the test suite for another app server but
without modifying anything except those parts of the test suite that are designed to be
modified

> Overall I'm not proposing anything radically new. I'm just proposing that specs who want to use more modern/familiar test frameworks such as Arquillian+JUnit can do so. The precedent has already been set by the CDI and Bean Validation specs (except they use TestNG instead of JUnit), and I find the CDI/BeanVal way of doing TCKs to be much easier to work with and maintain, so I'd like to do the same with JSON-B (and set an example for other specs too)

I understand.  No one likes JavaTest.  Creating a way to add additional tests using a new
framework without disrupting the existing JavaTest tests would be fine.  But it sounded like
you were converting existing tests to be run under the new framework.

aguibert:
> So in other words, it's not doing the same testing as the previous test, right?

No, the guts of the test code is still the same and the same assertions are being made -- just with different test framework. It is still doing the same testing as before -- I've taken the time to migrate the existing JSON-B TCKs to use Arquillian+JUnit instead of JavaTest.

You can see exactly what I changed to migrate from JT --> Arq+JUnit in this commit:
https://github.com/eclipse-ee4j/jsonb-api/pull/221/commits/62d17c980cae908ef4ff7539df25ddbacc36e2a9#diff-38839cd15ceb3c39fffd9e5723d54b5aR16

>  Creating a way to add additional tests using a new
framework without disrupting the existing JavaTest tests would be fine. But it sounded like
you were converting existing tests to be run under the new framework.

I am converting existing tests to run under a new framework. I think it's preferable for all TCK tests for a given spec to use the same framework. I suppose it's not a hard requirement, but I think it's preferable.

bshannon:
> > So in other words, it's not doing the same testing as the previous test, right?
> 
> No, the guts of the test code is still the same and the same assertions are being made -- just with different test framework. It is still doing the same testing as before -- I've taken the time to migrate the existing JSON-B TCKs to use Arquillian+JUnit instead of JavaTest.

If the test environment is different, it's testing a different thing.

> You can see exactly what I changed to migrate from JT --> Arq+JUnit in this commit:
> [eclipse-ee4j/jsonb-api@62d17c9#diff-38839cd15ceb3c39fffd9e5723d54b5aR16](https://github.com/eclipse-ee4j/jsonb-api/commit/62d17c980cae908ef4ff7539df25ddbacc36e2a9#diff-38839cd15ceb3c39fffd9e5723d54b5aR16)

That might help if I was deeply familiar with Arquillian, which I am not.

> > Creating a way to add additional tests using a new
> > framework without disrupting the existing JavaTest tests would be fine. But it sounded like
> > you were converting existing tests to be run under the new framework.
> 
> I am converting existing tests to run under a new framework. I think it's preferable for all TCK tests for a given spec to use the same framework. I suppose it's not a hard requirement, but I think it's preferable.

The problem is that you have to prove that your converted tests have all the same test behavior
as the original tests - that they pass *and fail* in all the same situations.  Do they run the same
tests in the EJB container, and in the web container, and (when appropriate) in the app client
container?

aguibert:
hi @bshannon 

> The problem is that you have to prove that your converted tests have all the same test behavior
as the original tests - that they pass and fail in all the same situations.

This seems rather unrealistic. I trust that this can be covered in a code review situation, as the modifications are quite straightforward.

> Do they run the same tests in the EJB container, and in the web container, and (when appropriate) in the app client
container?

In general yes -- Arquillian tests can run in any of these containers. For JSON-B specifically, the spec does not define any special interactions for container types -- it is essentially a standalone JavaSE library with some light CDI integration. Implementors _could_ re-run the tests in EJB container and app client if they wanted, but it would not provide any additional test value.

bshannon:
> hi @bshannon
> 
> > The problem is that you have to prove that your converted tests have all the same test behavior
> > as the original tests - that they pass and fail in all the same situations.
> 
> This seems rather unrealistic. I trust that this can be covered in a code review situation, as the modifications are quite straightforward.

If we're replacing a test suite that's used for conformance with a new test suite, we need
high confidence that the new test suite won't allow a product to pass that would've failed
with the old test suite.  If someone wants to set up a formal code review process to verify
that, that would be good.  But we need more than "ya, I think it should be the same".

> > Do they run the same tests in the EJB container, and in the web container, and (when appropriate) in the app client
> > container?
> 
> In general yes -- Arquillian tests can run in any of these containers. For JSON-B specifically, the spec does not define any special interactions for container types -- it is essentially a standalone JavaSE library with some light CDI integration. Implementors _could_ re-run the tests in EJB container and app client if they wanted, but it would not provide any additional test value.

If the current test suite runs the tests in all containers, the new test suite needs to as well,
whether or not you *think* the results would be the same.

aguibert:
>  If someone wants to set up a formal code review process to verify
that, that would be good. But we need more than "ya, I think it should be the same".

I previously referenced [this commit](https://github.com/eclipse-ee4j/jsonb-api/commit/62d17c980cae908ef4ff7539df25ddbacc36e2a9#diff-38839cd15ceb3c39fffd9e5723d54b5a) which both @m0mus and @jansupol reviewed. To me that seems like a sufficient review because we had another JSON-B SME and a TCK SME review the changes. What sort of formal review did you have in mind?

aguibert:
Closing this PR because @bshannon and I are in agreement that housing run configs for "externalized" TCKs is not the end result that we want.

Further discussions on this topic should continue on the jakartaee-platform-dev mailing list [on this thread](https://www.eclipse.org/lists/jakartaee-platform-dev/msg01455.html).

scottmarlow:
Hi @aguibert 

Can you please confirm that the JSONB API team is continuing to use the Standalone JSONB TCK that is generated from the Platform TCK?

I'm asking on this closed issue, as I would like to respond to your question ```What sort of formal review did you have in mind?```.

Thanks,
Scott

aguibert:
hi @scottmarlow, yes, in Yasson we are still using this approach as part of our CI pipeline. The full configuration can be found here:
https://github.com/eclipse-ee4j/yasson/blob/master/yasson-tck/pom.xml

Issue #163, opened 2020-03-13
com/sun/ts/tests/ejb32/lite/timer/schedule/expire/Client.java#leapYears failed due to wrong expected timeout


chengfang:
When running with WildFly, the following 4 tests failed due to the wrong expected next timeout value computed by the test:

```
com/sun/ts/tests/ejb32/lite/timer/schedule/expire/Client.java\#leapYears_from_ejblitejsf 
com/sun/ts/tests/ejb32/lite/timer/schedule/expire/Client.java\#leapYears_from_ejblitejsp 
com/sun/ts/tests/ejb32/lite/timer/schedule/expire/Client.java\#leapYears_from_ejbliteservlet 
com/sun/ts/tests/ejb32/lite/timer/schedule/expire/Client.java\#leapYears_from_ejbliteservlet2 
```

The relevant part of the test output shows that a timer is created to expire on Feb 29, 2028 **2:31:27**, and the test expects the next timeout to be Feb 29, 2028 **03:31:27**, but the actual next timeout is Feb 29 **02:31:27**.

The test timer has a time (Feb 29 02:31:27) that falls in US Standard Time, while the test execution date (Mar 12) is in US Daylight Saving Time, which is one hour ahead of US Standard Time. The test computes the expected timeout value based on the test execution time by setting year, month and day field, hence resulting in one hour ahead of the configured hour.

```
Mar 12, 2020 2:31:27 AM com.sun.ts.tests.common.vehicle.ejbliteshare.EJBLiteWebVehicleRunner run
INFO: Connecting http://localhost:8080/ejb32_lite_timer_schedule_expire_ejblitejsf_vehicle_web/faces/ejblitejsf_vehicle.jsp?testName=leapYears
<html>
    <head><title>leapYears_from_ejblitejsf</title></head>
    <body>
        [TEST FAILED] testName=leapYears 
Created a timer with expression year=2025-2028 month=2 dayOfMonth=29, 29 dayOfWeek=* hour=2 minute=31 second=27 start=null end=null timezone=null
Compare expected nextTimeout Tue Feb 29 03:31:27 EST 2028, and actual nextTimeout Tue Feb 29 02:31:27 EST 2028
```

bshannon:
Can you submit a Pull Request to fix these tests?

chengfang:
Sure, I'll give it a try.

scottmarlow:
Cheng, thanks for the fix, since your change is merged now to master, I will close this issue.

PR #165, opened 2020-03-16
Issue #163 .../ejb32/lite/timer/schedule/expire/Client.java#leapYears failed due to wrong expected timeout

chengfang:
com/sun/ts/tests/ejb32/lite/timer/schedule/expire/Client.java#leapYears failed due to wrong expected timeout

Signed-off-by: Cheng Fang <cfang@redhat.com>

### DIFF
--- a/src/com/sun/ts/tests/ejb32/lite/timer/schedule/expire/Client.java
+++ b/src/com/sun/ts/tests/ejb32/lite/timer/schedule/expire/Client.java
@@ -31,10 +31,16 @@ import java.util.Date;
 import java.util.Properties;
 
 /**
- * Some tests use 2100 as a test calendar: January February Su Mo Tu We Th Fr Sa
- * Su Mo Tu We Th Fr Sa 1 2 1 2 3 4 5 6 3 4 5 6 7 8 9 7 8 9 10 11 12 13 10 11 12
- * 13 14 15 16 14 15 16 17 18 19 20 17 18 19 20 21 22 23 21 22 23 24 25 26 27 24
- * 25 26 27 28 29 30 28 31
+ * Some tests use 2100 as a test calendar:
+       January               February
+ Su Mo Tu We Th Fr Sa  Su Mo Tu We Th Fr Sa
+                 1  2      1  2  3  4  5  6
+  3  4  5  6  7  8  9   7  8  9 10 11 12 13
+ 10 11 12 13 14 15 16  14 15 16 17 18 19 20
+ 17 18 19 20 21 22 23  21 22 23 24 25 26 27
+ 24 25 26 27 28 29 30  28
+ 31
+
  * 
  */
 public class Client extends ClientBase {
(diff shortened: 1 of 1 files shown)


src/com/sun/ts/tests/ejb32/lite/timer/schedule/expire/Client.java
```diff
@@ -463,7 +468,13 @@ public void leapYears() {
...
           Calendar.FEBRUARY);
       Timer timer = createTimer(exp);
-      verifyNextTimeout(expectedNextTimeout, timer);
+
+      // Use custom ignorable millis to accommodate the hour difference between
+      // the test timer's date (e.g., Feb 29) and execution date (e.g., March)
+      // due to daylight time switch in certain regions.
+      // See https://github.com/eclipse-ee4j/jakartaee-tck/issues/163
+      verifyNextTimeout(expectedNextTimeout, timer,
+              DateUtils.MILLIS_PER_HOUR + DateUtils.MILLIS_PER_MINUTE);
```
scottmarlow: @chengfang would it be better to change the tests to be consistent about either using US Standard Time or US Daylight Saving Time, but not both?

markt-asf: Better still, use UTC.

chengfang: These tests are testing leap years, so my thinking was to verify year, month and day, while be loose on hours and minutes, without complicating the tests with timezone. If we want to add a consistent timezone to this set of tests, I can give modify the fix and test how it goes.

chengfang:
modified the fix to use UTC in both test timers and computed expected timeout value, to avoid the hour diff involving daylight time switch.

src/com/sun/ts/tests/ejb32/lite/timer/schedule/expire/Client.java
```diff
@@ -449,19 +462,17 @@ public void leapYears() {
...
       ScheduleExpression exp = TimerUtil.getPreciseScheduleExpression(cal)
-          .dayOfMonth(dayOfMonths[i]).month(2).year(yearRange[i]);
-
-      Date expectedNextTimeout = TimerUtil.getCurrentDatePlus(0, 0);
-      expectedNextTimeout = DateUtils.setYears(expectedNextTimeout,
-          leapYears[i]);
-      expectedNextTimeout = DateUtils.setDays(expectedNextTimeout, 29);
-      expectedNextTimeout = DateUtils.setMonths(expectedNextTimeout,
-          Calendar.FEBRUARY);
+          .dayOfMonth(dayOfMonths[i]).month(2).year(yearRange[i]).timezone(DateUtils.UTC_TIME_ZONE.getID());
```
bshannon: It would be better if this and the lines above stuck within the 80 column line width
used by the original code.

Otherwise, looks good to me, but get an EJB expert to look at it too!

scottmarlow:
looks good to me as well, I'll merge it and we will see the TCK results tonight.

scottmarlow:
If it causes problems, we can revert later.  I'd also like to merge this before any package name changes get merged (seems like we are not ready to merge those changes yet).

PR #161, opened 2020-03-10
batch property name + switch to jakarta package changes

scottmarlow:
This pr only has three package name changes.  It is the result of applying the following commands:

> find -name "*.java" -type f| xargs sed -i 's/javax\.annotation\.security\./jakarta\.annotation\.security\./g'
> find -name "*.java" -type f| xargs sed -i 's/javax\.annotation\.sql\./jakarta\.annotation\.sql\./g'
> find -name "*.java" -type f| xargs sed -i 's/javax\.batch\./jakarta\.batch\./g'

And committing the changed sources as separate commits.

Is anyone else already doing this?  

Does anyone know which files in the platform TCK need to be updated to bring in the EE 9 SPEC API artifacts?

PR #167, opened 2020-03-18
convert end of lines in src/com/ibm/jbatch/tck/testJobXml/META-INF/batch-jobs

scottmarlow:

Was previously part of [pull#161](https://github.com/eclipse-ee4j/jakartaee-tck/pull/161), moved here so the end of line changes can be merged in later after other changes are made (any other batch changes that precede the namespace change or after we switch to jakarta namespace).


scottmarlow:
I believe that this pr is no longer needed as I think another pr took care of this.

Issue #125, opened 2019-09-12
ejb32/lite/timer/schedule/expression/annotated/Client.java\#dayOfWeekOverDayOfMonth failed at midnight


chengfang:
This issue affects the following tests:

```
com/sun/ts/tests/ejb32/lite/timer/schedule/expression/annotated/Client.java\#dayOfWeekOverDayOfMonth_from_ejbliteservlet2

com/sun/ts/tests/ejb32/lite/timer/schedule/expression/annotated/Client.java\#dayOfMonthOverDayOfWeek_from_ejbliteservlet2

com/sun/ts/tests/ejb32/lite/timer/schedule/expression/annotated/Client.java\#dayOfMonthOverDayOfWeek_from_ejbliteservlet 

com/sun/ts/tests/ejb32/lite/timer/schedule/expression/annotated/Client.java\#dayOfMonthOverDayOfWeek_from_ejblitejsf

com/sun/ts/tests/ejb32/lite/timer/schedule/expression/annotated/Client.java\#dayOfMonthOverDayOfWeek_from_ejblitejsp

com/sun/ts/tests/ejb32/lite/timer/schedule/expression/annotated/Client.java\#dayOfWeekOverDayOfMonth_from_ejblitejsf 

com/sun/ts/tests/ejb32/lite/timer/schedule/expression/annotated/Client.java\#dayOfWeekOverDayOfMonth_from_ejblitejsp 
```

similar tests under com/sun/ts/tests/ejb30/timer/...

It failed because the test ran at mid-night (11:59:59 PM).  The test created a timer to fire the next day, and expect no timer is fired.  When the timer is created at the end of day 1, but it was already day 2 when checking test results.

Snippet from test log:

```
Sep 03, 2019 11:59:59 PM com.sun.ts.tests.common.vehicle.ejbliteshare.EJBLiteWebVehicleRunner run

[TEST FAILED] testName=dayOfWeekOverDayOfMonth Created timer: TimerInfo
testName=dayOfWeekOverDayOfMonth testNumber=0 status=false doubleVar=0.0 longVar=0 charVar=  booleanVar=false intVar=0 schedule:ScheduleExpression [second=*;minute=*;hour=*;dayOfMonth=*;month=*;dayOfWeek=3;year=*;timezoneID=null;start=null;end=null] nextTimeout:Wed Sep 04 00:00:00 EDT 2019 timeRemaining:130 isPersistent:false
```

```
[...]

PR #171, opened 2020-03-22
Fix issue #125 timer/schedule/expression/annotated/Client.java#dayOfW…

chengfang:
…eekOverDayOfMonth failed at midnight

Signed-off-by: Cheng Fang <cfang@redhat.com>

### DIFF
--- a/src/com/sun/ts/tests/ejb30/timer/schedule/expression/annotated/ScheduleBean.java
+++ b/src/com/sun/ts/tests/ejb30/timer/schedule/expression/annotated/ScheduleBean.java
@@ -217,10 +217,10 @@ public class ScheduleBean extends TimerBeanBaseWithoutTimeOutMethod {
   public String dayOfWeekOverDayOfMonth() {
     TimerInfo info = new TimerInfo("dayOfWeekOverDayOfMonth");
     Calendar cal = Calendar.getInstance();
-    cal.add(Calendar.DAY_OF_WEEK, 1);
+    cal.add(Calendar.DAY_OF_WEEK, 2);
     int nextDayOfWeek = TimerUtil.getForSchedule(Calendar.DAY_OF_WEEK, cal);
     // every year, every month, every hour, every minute, every second, but not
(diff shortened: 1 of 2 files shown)


src/com/sun/ts/tests/ejb30/timer/schedule/expression/annotated/ScheduleBean.java
```diff
@@ -217,10 +217,10 @@ public String attributeDefaults() {
   public String dayOfWeekOverDayOfMonth() {
     TimerInfo info = new TimerInfo("dayOfWeekOverDayOfMonth");
     Calendar cal = Calendar.getInstance();
-    cal.add(Calendar.DAY_OF_WEEK, 1);
+    cal.add(Calendar.DAY_OF_WEEK, 2);
     int nextDayOfWeek = TimerUtil.getForSchedule(Calendar.DAY_OF_WEEK, cal);
     // every year, every month, every hour, every minute, every second, but not
-    // every day. Only for the next day of week. So the timeout method for
+    // every day. Only for 2+ day of week. So the timeout method for
```
scottmarlow: What is wrong currently?

Why do we need to make this change?

Thanks,
Scott

chengfang:
The test creates a timer that expires any time next day, and expects there will be no timeout event. When it is run near midnight (11:59:xx PM), the next day is only a second or less away. By the time the test checks for any timeout, the timer already fired since it's already the next day.

So my fix is to change the next day to next next day to make sure it will not expire as expected by the test.  See https://github.com/eclipse-ee4j/jakartaee-tck/issues/125 for more details, and logs showing the failed execution time is always 11:59:xx PM.

scottmarlow:
Awesome, thanks for the explanation Cheng!  :)

scottmarlow:
After we have converted the Jakarta EE TCK master branch to use the Jakarta namespace, we can merge this change.

scottmarlow:
Thanks @chengfang

PR #174, opened 2020-03-26
convert non-jdk javax.transaction => jakarta.transaction

scottmarlow:


scottmarlow:
Looks like this is needed to get further in the build, as during [today's jakartaee-tck/job/master build](https://ci.eclipse.org/jakartaee-tck/job/jakartaee-tck/job/master/539/) we get:

`
 [ts.javac] /home/jenkins/agent/workspace/jakartaee-tck_master/src/com/sun/ts/tests/compat13/ejb/tx/TestBeanEJB.java:139: error: incompatible types: jakarta.transaction.UserTransaction cannot be converted to javax.transaction.UserTransaction
 [ts.javac]       ut = sctx.getUserTransaction();
 [ts.javac]                                   ^
`

scottmarlow:
I need to update the Copyright dates, to follow https://www.eclipse.org/projects/handbook/#ip-copyright-headers, as instead of "Copyright (c) 2018, 2019-2020", we should instead use "Copyright (c) 2018, 2020"

scottmarlow:
The copyright dates should now match the Eclipse (optional) recommendation as mentioned above.

scottmarlow:
Requested a few more reviewer, if one of you could review/approve, will merge this.

scottmarlow:
note that https://github.com/eclipse-ee4j/jakartaee-tck/pull/182 also updates the javax.transaction references.

alwin-joseph:
> I need to update the Copyright dates, to follow https://www.eclipse.org/projects/handbook/#ip-copyright-headers, as instead of "Copyright (c) 2018, 2019-2020", we should instead use "Copyright (c) 2018, 2020"
> 
> Or, we could simply stick with 2018. As per text quoted from above link:
> 
> `The {year} is the year in which the content was created (e.g. “2004”). The date may optionally be set to a range of years with the first and last years of the range separated by a comma (e.g. “2004, 2017”); in this case, the first year is when the content was first created and the last year is when the content was last modified.`

Yes we had followed the "first and last years" in our previous changes. e.g. “2004, 2017”.  
Do you know if the Eclipse (optional) recommendation allows 3 years as in few files. 

Also the signature-repository/javax* files would be removed once we have the correct jakarta* sig files.

scottmarlow:
Thanks, I will fix the sources with 3 years in copyright, that is not allowed by the Eclipse (optional) recommendation.

Should we revert the signaturetest changes?  As like you said, they really need to regenerated, so it might be less confusing if we don't manually update the javax signature files.

alwin-joseph:
> @alwin-joseph
> 
> Thanks, I will fix the sources with 3 years in copyright, that is not allowed by the Eclipse (optional) recommendation.

Those are the javax* signature files only as I see.
> 
> Should we revert the signaturetest changes? As like you said, they really need to regenerated, so it might be less confusing if we don't manually update the javax signature files.

The changes are already made to start using jakarta* sig files , even though the sig files need correction. So I think it better be removed from the PR to avoid confusion.

scottmarlow:
thanks @alwin-joseph, signature files are also removed.  Files with 3 dates are now updated to have 2 dates.

PR #175, opened 2020-03-26
convert javax.el => jakarta.el

scottmarlow:


install/jsf/docs/assertions/JSFJavadocAssertions.html
```diff
@@ -3,7 +3,7 @@
 <META http-equiv="Content-Type" content="text/html; charset=UTF-8">
 <!--
 
-    Copyright (c) 2018, 2019 Oracle and/or its affiliates. All rights reserved.
+    Copyright (c) 2018, 2019-2020 Oracle and/or its affiliates. All rights reserved.
```
kwsutter: The copyright date format should be `"<beginning date>, <ending date>"`.  So, this update should be "2018, 2020".

I'm starting a conversation on the PMC list as to whether the Oracle attribution should be updated or not...

scottmarlow: I manually made the date changes.

I'm using [copyright-year-updater.sh that I found online](https://github.com/fwolf/copyright-year-updater.sh), that claims to follow the GNU convention documented on [https://www.gnu.org/prep/maintain/html_node/Copyright-Notices.html](https://www.gnu.org/prep/maintain/html_node/Copyright-Notices.html).  It doesn't look like the script handled changing "2018, 2019" into "2018-2020".  Will have to look for a better way to update the TCK copyright dates, or create a pr for copyright-year-updater.sh.

It looks like copyright-year-updater.sh did okay with some of the EJB changes, like, in the following, no range was added (above gnu link mentions that range is only valid when significant changes are made in those years, whatever that means):
`
src/com/sun/ts/tests/xa/ee/resXcomp2/Ejb2TestHome.java: * Copyright (c) 2007, 2018, 2020 Oracle and/or its affiliates. All rights reserved.
`

kwsutter: @scottmarlow Yeah, I was having a discussion with Steve Millidge about the "significant changes" idea.  His thought is that the updates for the `jakarta.*` namespace change do not warrant an update to the "contributor portion" of the Copyright.  That is, it's not a significant feature update or significant thought required.  But, the date should be still be updated regardless.

kwsutter: Also, here's a link to the Eclipse Handbook's information on the use of Copyrights.  That should be the reference used, not necessarily the gnu reference.  https://www.eclipse.org/projects/handbook/#ip-copyright-headers

scottmarlow: Thanks Kevin, agreed, we should follow https://www.eclipse.org/projects/handbook/#ip-copyright-headers

scottmarlow: Quoting from the linked description:
`
The {year} is the year in which the content was created (e.g. “2004”). The date may optionally be set to a range of years with the first and last years of the range separated by a comma (e.g. “2004, 2017”); in this case, the first year is when the content was first created and the last year is when the content was last modified.
`
We could use either "Copyright (c) 2007" or "Copyright (c) 2007, 2020", however, I do like the optional date format better.

Looks like we have some dates to also correct on [the javax.transaction pr](https://github.com/eclipse-ee4j/jakartaee-tck/pull/174) as well.

bshannon: > I manually made the date changes.
> 
> I'm using [copyright-year-updater.sh that I found online](https://github.com/fwolf/copyright-year-updater.sh), 

I guess you're not aware of the [GlassFish copyright tool](https://github.com/eclipse-ee4j/glassfish-copyright-plugin/blob/master/README.md).  It's full of customizations for these projects.

bshannon: > Quoting from the linked description:
> `The {year} is the year in which the content was created (e.g. “2004”). The date may optionally be set to a range of years with the first and last years of the range separated by a comma (e.g. “2004, 2017”); in this case, the first year is when the content was first created and the last year is when the content was last modified.`
> We could use either "Copyright (c) 2007" or "Copyright (c) 2007, 2020", however, I do like the optional date format better.

At this point it would be more confusing to change the conventions we've been using
for years, especially since you can't tell by looking at the file which convention it is
following.

Plus, lawyers change their minds on this stuff far too often.  It's better to have more
information there than less.

PR #183, opened 2020-04-01
Temporary fix to use jakartaee-api.jar to build standalone TCKs

alwin-joseph:


alwin-joseph:
@scottmarlow FYI, I have made this change to use jakartaee-api.jar in classpath to allow standalone TCK build.

PR #188, opened 2020-04-02
JAX-RPC is deprecated in Jakarta EE 9, so remove call to SessionContext#getMessageContext()

scottmarlow:


scottmarlow:
don't bother reviewing this

PR #197, opened 2020-04-12
Web sock tck pkg rename

frowe:
Updated pkg names from javax -> jakarta.  Because the directory hierarchy of the tests had things like ee/javax, renamed dir paths also.


frowe:
Ignore this PR for now, I've got to correct it.

frowe:
Replaced with #202

PR #199, opened 2020-04-13
jsonb TCK package rename javax->jakarta

frowe:
change package javax->jakarta


scottmarlow:
Looks good, thanks @frowe!

PR #215, opened 2020-04-18
Increase memory size for Jenkins JNLP process from default, up to 2176MB (2GB + 128MB)

scottmarlow:
Experiment to set the memory size for the Jenkins JNLP process to be a little larger than the max Java heap size specified (-Xmx2048m).  This may help us avoid git failures that we see sometimes.

Changed JNLP config from:
```
  - name: jnlp
    env:
      - name: JNLP_PROTOCOL_OPTS
        value: "-XshowSettings:vm -Xmx2048m -Dsun.zip.disableMemoryMapping=true -Dorg.jenkinsci.remoting.engine.JnlpProtocol3.disabled=true"
```

To:
```
  - name: jnlp
    env:
      - name: JNLP_PROTOCOL_OPTS
        value: "-XshowSettings:vm -Xmx2048m -Dsun.zip.disableMemoryMapping=true -Dorg.jenkinsci.remoting.engine.JnlpProtocol3.disabled=true"
    resources:
      limits:
        memory: 2176Mi
```

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -128,6 +128,9 @@ spec:
     env:
       - name: JNLP_PROTOCOL_OPTS
         value: "-XshowSettings:vm -Xmx2048m -Dsun.zip.disableMemoryMapping=true -Dorg.jenkinsci.remoting.engine.JnlpProtocol3.disabled=true"
+    resources:
+      limits:
+        memory: 2176Mi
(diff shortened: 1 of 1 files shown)


scottmarlow:
I'm merging this, will revert if it causes problems.

PR #252, opened 2020-05-11
Update documentation for sigTestClasspath - add Java 11 info

markt-asf:
This is a pure documentation fix. I have only updated the ts.jte comments for EL as that is what I am currently testing (Tomcat 10 is passing the nightly build for the EL TCK on Java 8 and Java 11 so all looking good). I'm happy to expand this to update all the ts.jte files if that would be useful. Just leave a review comment to that effect.

Another option is the include both `${java.home}/lib/rt.jar` and `${java.home}/lib/modules` in the default value for `sigTestClasspath` so it "just works" for Java 8 and Java 11 on most platforms.


scottmarlow:
Looks good to me, thanks Mark!

Issue #234, opened 2020-05-05
javax.annotation => jakarta.annotation


scottmarlow:
Rename javax.annotation => jakarta.annotation as needed.
Also figure out why Jakarta SPEC JAR + JDK have some common classes:

Output from jar tf /tmp/jakarta.jakartaee-api-9.0.0-RC1.jar | grep jakarta.annotation:
```
jakarta/annotation/
jakarta/annotation/security/
jakarta/annotation/sql/
jakarta/annotation/PreDestroy.class
jakarta/annotation/security/DenyAll.class
jakarta/annotation/security/PermitAll.class
jakarta/annotation/security/RunAs.class
jakarta/annotation/security/DeclareRoles.class
jakarta/annotation/security/RolesAllowed.class
jakarta/annotation/Priority.class
jakarta/annotation/Resource$AuthenticationType.class
jakarta/annotation/Resource.class
jakarta/annotation/sql/DataSourceDefinition.class
jakarta/annotation/sql/DataSourceDefinitions.class
jakarta/annotation/PostConstruct.class
jakarta/annotation/Resources.class
jakarta/annotation/Generated.class
jakarta/annotation/ManagedBean.class
```

jdk has: 
```
javax.annotation.Generated
javax.annotation.PostConstruct
javax.annotation.PreDestroy
javax.annotation.Resource
javax.annotation.Resources
```

scottmarlow:
https://github.com/javaee/javax.annotation/tree/master/src/main/java/javax/annotation also has the same, giving that was in Java EE 8, it must be right, so I no longer have a question about the overlap between SE + EE.

scottmarlow:
Thanks @RohitKumarJain

PR #255, opened 2020-05-12
Renaming javax.annotation => jakarta.annotation

RohitKumarJain:
Creating this PR to address https://github.com/eclipse-ee4j/jakartaee-tck/issues/234


scottmarlow:
@RohitKumarJain I think that we need to understand why the GlassFish jakarta.jakartaee-api-9.0.0-RC1.jar contains annotation classes that are also in the JDK.  I added a list of classes in [issues/234](https://github.com/eclipse-ee4j/jakartaee-tck/issues/234).  What do you think?

scottmarlow:
As mentioned elsewhere, https://github.com/javaee/javax.annotation/tree/master/src/main/java/javax/annotation seems to also have the same annotation, so I don't think it was an oversight.

Issue #219, opened 2020-04-23
Update signature tests for JDK 8/11


scottmarlow:
[Follow instructions Signature-Tests-in-JakartaEE-TCK-for-JDK11 instructions for JDK 11](https://github.com/eclipse-ee4j/jakartaee-tck/wiki/Signature-Tests-in-JakartaEE-TCK-for-JDK11).

Also consider [updating signaturetest README for issue#181](https://github.com/eclipse-ee4j/jakartaee-tck/issues/181) when updating signatures for JDK 8 as well as for JDK 11.

scottkurz:
I would like to understand this area better since Batch has its own standalone TCK that I'd also like to produce the signature tests for, in addition to helping, if possible, with the platform TCK.

I thought a simple first step would be to just try running the Java 11 Batch sig tests against the platform milestone API JAR and eventually the batch CI... however I was unable to get this to work.
I'm not clear how to run the tests.

If I take the  **jakarta.batch.sig_1.0_se11** committed to the repo now and run (Java 11):

```bash
REPO_ROOT= # .. git clone here
cd $REPO_ROOT
java -jar $REPO_ROOT/lib/sigtest.jar   SignatureTest -static -package jakarta.batch -filename   src/com/sun/ts/tests/signaturetest/signature-repository/jakarta.batch.sig_1.0_se11   -classpath  jakarta.jakartaee-api-9.0.0-RC1.jar
```

it fails with 39 errors.  This gives me a clue that it's pretty clear that the problem is that the java.lang.* classes are not on the classpath.  I say that because it's the Exception classes and the abstract classes (extending java.lang.Object) which result in failure.     

OK, so I look at these instructions:
https://github.com/eclipse-ee4j/jakartaee-tck/wiki/Signature-Tests-in-JakartaEE-TCK-for-JDK11

and it suggests I should be using `jimage` to extract the modules, such as `java.base`, which should be on my classpath.

However, I've tried a bunch of variants of classpaths and have tried Open JDK and Oracle JDK, but can not build a classpath and get it to work.   Everything I try results in this unhelpful error message:

`STATUS:Error.Class path not specified or incorrect in the static mode.`

with no clue what the problem is or the next step.

True, there a couple READMEs to look at but also warnings with issues like: https://github.com/eclipse-ee4j/jakartaee-tck/issues/181  saying these may be outdated.

Appreciate any advice on the next step.  Thanks,

PR #256, opened 2020-05-12
WebSocket signature update

markt-asf:
The WebSocket team have made a small, backwards compatible change to the WebSocket API to add default methods where appropriate. This PR updates the signature tests for these changes.

### DIFF
--- a/src/com/sun/ts/tests/signaturetest/signature-repository/jakarta.websocket.sig_2.0_se11
+++ b/src/com/sun/ts/tests/signaturetest/signature-repository/jakarta.websocket.sig_2.0_se11
@@ -219,8 +219,8 @@ innr public abstract interface static Binary
 innr public abstract interface static BinaryStream
 innr public abstract interface static Text
 innr public abstract interface static TextStream
-meth public abstract void destroy()
-meth public abstract void init(jakarta.websocket.EndpointConfig)
+meth public void destroy()
+meth public void init(jakarta.websocket.EndpointConfig)
 
 CLSS public abstract interface static jakarta.websocket.Decoder$Binary<%0 extends java.lang.Object>
  outer jakarta.websocket.Decoder
@@ -261,8 +261,8 @@ innr public abstract interface static Binary
 innr public abstract interface static BinaryStream
 innr public abstract interface static Text
 innr public abstract interface static TextStream
-meth public abstract void destroy()
-meth public abstract void init(jakarta.websocket.EndpointConfig)
+meth public void destroy()
+meth public void init(jakarta.websocket.EndpointConfig)
 
 CLSS public abstract interface static jakarta.websocket.Encoder$Binary<%0 extends java.lang.Object>
  outer jakarta.websocket.Encoder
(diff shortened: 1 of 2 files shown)


scottmarlow:
Hi @markt-asf my understanding is that the signature files need to be generated, instead of edited directly.  This would be after pulling in the various Jakarta EE 9 SPEC APIs (including WebSocket 2.0 API).  The README for the process is in [JDK8 README](https://github.com/eclipse-ee4j/jakartaee-tck/tree/master/src/com/sun/ts/tests/signaturetest) + [JDK11 README](https://github.com/eclipse-ee4j/jakartaee-tck/wiki/Signature-Tests-in-JakartaEE-TCK-for-JDK11).  

Tracking issue for [updating the signature files is here.](https://github.com/eclipse-ee4j/jakartaee-tck/issues/219) (assigned to Alwin Joseph).

markt-asf:
The current Java 11 signature file for WebSocket was not generated, it was hand edited from the Java 8 file. I know because I did the work.
I don't understand the reluctance to apply a known good fix to a known issue that would get us a step closer (and all the way once the other PRs are applied) to having a WebSocket TCK for Jakarta EE 9 that can be passed by a compliant implementation.
I also question the process of simply generating the signature files from the API. There is a risk that inadvertent changes to the API will get incorporated in the signature files. Manual changes to the signature files followed by validation that the automated process generates the same result would look to be a safer approach.

scottmarlow:
We had some discussion previously about what is allowed with regard to API changes.  I forget if that was on the Platform mailing list or elsewhere.  I don't recall anyone asking about removing abstract keyword from an API method but maybe that is okay.  Is it okay if we find that previous discussion and ask about removing "abstract" from a SPEC API there first?

Regarding whether the API signatures are always generated or not from the released SPEC API, is a different question to answer.  I do see that you did generate those before as part of commit 510b54058ee057a0b19d3136a89c2cea6bc6ef4a (javax -> jakarta for WebSocket TCK).

So, yes, I agree that merging your changes would be fine (since that is the same as we did earlier this year) but keep in mind that when we do soon generate the signatures, the generated output will be used.  If there are differences for WebSocket, that would be surprising from what you said, we can discuss later if that is the case.

scottmarlow:
> I also question the process of simply generating the signature files from the API. There is a risk that inadvertent changes to the API will get incorporated in the signature files. Manual changes to the signature files followed by validation that the automated process generates the same result would look to be a safer approach.

Is the root issue that the signature tool shouldn't be generating abstract on interface methods?

markt-asf:
No. My point is that it is possible for someone to inadvertently change the API in a non-backwards compatible way. Normally this would get caught during review but it might not. If a tool is then used to automatically update the signature files then the API change will, unintentionally, be made part of the TCK.

My suggested alternative approach is to manually update the signatures so any changes are deliberate and then use to tool to ensure that the manually updated signature files are consistent with the output of the tool when run against the API. Any discrepancies will be highlighted and can then be investigated.

In the case of this particular PR, the WebSocket community has made a deliberate decision to make a 100% backwards compatible change to the API (adding a few default methods).

Issue #254, opened 2020-05-12
Typo and wrong test  in servlet tck


olamy:
There is a typo here https://github.com/eclipse-ee4j/jakartaee-tck/blob/a50bf877a5413b812fad569d116c38160ba0aa41/src/com/sun/ts/tests/servlet/spec/errorpage/SecondServletErrorPage.java#L40 `javax.serlvet.error.exception_type` rather than `javax.servlet.error.exception_type` so the test assert the value is null which is wrong because it should contains the class name of the exception. Test here: https://github.com/eclipse-ee4j/jakartaee-tck/blob/a50bf877a5413b812fad569d116c38160ba0aa41/src/com/sun/ts/tests/servlet/spec/errorpage/URLClient.java#L149

scottmarlow:
In summary, the impacted test is currently ignored in all 8.0.x releases of the Jakarta EE Platform TCK, which hides a problem that @olamy is reporting via this issue.  @olamy will submit a pull request that corrects the typo, so that the test will actually run.  @olamy still needs to explain why the test (after typo is corrected) is not Servlet SPEC compliant and what the correction will be (perhaps via pull request).

Sound right @olamy ?

olamy:
@scottmarlow did you look at the PR #277 ?

scottmarlow:
@olamy does [PR#277](https://github.com/eclipse-ee4j/jakartaee-tck/pull/277) address the test failure as well?  Or does it only allow the test failure to occur?

olamy:
#277 is fixing the TCK. This means fixing the servlet request attribute name typo and so fix the corresponding assert/test

scottmarlow:
@olamy I thought that you said before that the test would fail after correcting the typo.  Ah, I see, so it is removing the "null" from what we are testing for.  Previously, the null, masked the fact that there is a typo "javax.serlvet.error.exception_type". :)

scottmarlow:
Closing as I think this is resolved now.  Please reopen if not.

PR #258, opened 2020-05-13
Issue #254 Fix typo in servlet attribute name containing exception type (fix for javax.servlet TCK)

olamy:
Currently the tests are using a constant with a typo `javax.serlvet.error.exception_type` note the `serlvet`  https://github.com/eclipse-ee4j/jakartaee-tck/pull/258/commits/b7360eee2dc21fa33a7a692bf44916cecdfb7990#diff-f09eab37c700ed051ff72ff5e9a13963L42 so the test assert the attribute with the typo in the name returns Null https://github.com/eclipse-ee4j/jakartaee-tck/pull/258/commits/b7360eee2dc21fa33a7a692bf44916cecdfb7990#diff-050a4404eb79e87a2927e2e5f722de06L75 whereas as the specs says this attribute should contains the class of the exception in the format `java.lang.Class`. 
So I changed the  constant to use constants from the servlet api jar class `RequestDispatcher` and I fix the assert as well to contains the class in the toString format.

### DIFF
--- a/src/com/sun/ts/tests/servlet/spec/errorpage/SecondServletErrorPage.java
+++ b/src/com/sun/ts/tests/servlet/spec/errorpage/SecondServletErrorPage.java
@@ -22,6 +22,7 @@ package com.sun.ts.tests.servlet.spec.errorpage;
 
 import com.sun.ts.tests.servlet.common.util.Data;
 
+import javax.servlet.RequestDispatcher;
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
@@ -35,17 +36,17 @@ import java.io.PrintWriter;
 
 public class SecondServletErrorPage extends HttpServlet {
 
-  private static final String STATUS_CODE = "javax.servlet.error.status_code";
+  private static final String STATUS_CODE = RequestDispatcher.ERROR_STATUS_CODE; // "javax.servlet.error.status_code";
 
-  private static final String EXCEPTION_TYPE = "javax.serlvet.error.exception_type";
+  private static final String EXCEPTION_TYPE = RequestDispatcher.ERROR_EXCEPTION_TYPE; //"javax.servlet.error.exception_type";
 
-  private static final String MESSAGE = "javax.servlet.error.message";
+  private static final String MESSAGE = RequestDispatcher.ERROR_MESSAGE; //"javax.servlet.error.message";
 
-  private static final String EXCEPTION = "javax.servlet.error.exception";
+  private static final String EXCEPTION = RequestDispatcher.ERROR_EXCEPTION;  //"javax.servlet.error.exception";
 
-  private static final String REQUEST_URI = "javax.servlet.error.request_uri";
+  private static final String REQUEST_URI = RequestDispatcher.ERROR_REQUEST_URI; //"javax.servlet.error.request_uri";
 
-  private static final String SERVLET_NAME = "javax.servlet.error.servlet_name";
+  private static final String SERVLET_NAME = RequestDispatcher.ERROR_SERVLET_NAME; //"javax.servlet.error.servlet_name";
 
   private static final String EXP_MESSAGE = "error page invoked";
 
@@ -60,7 +61,8 @@ public class SecondServletErrorPage extends HttpServlet {
     pw.println("Servlet Name: " + req.getAttribute(SERVLET_NAME));
     pw.println("Request URI: " + req.getAttribute(REQUEST_URI));
     pw.println("Status Code: " + req.getAttribute(STATUS_CODE));
-    pw.println("Exception Type: " + req.getAttribute(EXCEPTION_TYPE));
+    // with the cast we even enforce it's a Class type
+    pw.println("Exception Type: " + ((Class)req.getAttribute(EXCEPTION_TYPE)).getName());
     pw.println("Exception: " + req.getAttribute(EXCEPTION));
     pw.print("Message: ");
     if (((String) req.getAttribute(MESSAGE)).indexOf(EXP_MESSAGE) > -1) {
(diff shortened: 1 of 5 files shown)


alwin-joseph:
@olamy Thanks for the change. The issue looks like a bug to me, I will try to verify it more.

IMO such changes can also be merged to master branch. The 8.0 branch doesn't seem to be used  for any releases.  think this 8.0 branch can be deleted also @scottmarlow . For the past releases we were only maintaining tags 8.0.0, 8.0.1, 8.0.2 .

olamy:
ah ok fair enough I'm not sure about the process with tags/branches/master etc.. here (usually for me a tag is immutable so that's why I propose the change against 8.x branch) 
Let me know what is the best for you

kwsutter:
> @scottmarlow IMO such changes can also be merged to master branch. The 8.0 branch doesn't seem to be used for any releases. I think this 8.0 branch can be deleted. For the past releases we were only maintaining tags 8.0.0, 8.0.1, 8.0.2 .

I agree that the 8.0 branch is confusing.  But, if we're maintaining the 8.0 stream for the 8.0.0, 8.0.1, 8.0.2, 8.0.x releases, then you do need some branch separate from master.  Otherwise, if an update is required for 8.0.3 (as an example), you don't have a branch that represents that code.  Maybe we should rename the 8.0 branch to 8.0.x for maintenance purposes.  Or, if we don't know the history of that 8.0 branch, maybe a new branch should be created at the 8.0.2 tag and call it 8.0.x.  Or, just delete the 8.0 branch and create a new branch if or when it's necessary.  We need to do something so that contributors like @olamy don't accidentally try to update the wrong branch...

scottmarlow:
> 
> 
>     @scottmarlow IMO such changes can also be merged to master branch. The 8.0 branch doesn't seem to be used for any releases. I think this 8.0 branch can be deleted. For the past releases we were only maintaining tags 8.0.0, 8.0.1, 8.0.2 .
> 
> I agree that the 8.0 branch is confusing. But, if we're maintaining the 8.0 stream for the 8.0.0, 8.0.1, 8.0.2, 8.0.x releases, then you do need some branch separate from master. Otherwise, if an update is required for 8.0.3 (as an example), you don't have a branch that represents that code. Maybe we should rename the 8.0 branch to 8.0.x for maintenance purposes. Or, if we don't know the history of that 8.0 branch, maybe a new branch should be created at the 8.0.2 tag and call it 8.0.x. Or, just delete the 8.0 branch and create a new branch if or when it's necessary. We need to do something so that contributors like @olamy don't accidentally try to update the wrong branch...

IMO, the only 8.0.x activity expected, would be excluding challenged tests (following the TCK challenge process).  I agree that we don't need the 8.0 branch and agree with deleting the branch, to avoid confusion.  If we ever need a 8.something branch, that could be created off of the latest 8.0.n tag (like suggested above).

alwin-joseph:
@olamy Were you able to get the tests passing with this change.

I am getting below 3 test failures when run with glassfish. I did not investigate further.

com.sun.ts.tests.servlet.spec.errorpage.URLClient.heirarchyErrorMatchTest
com.sun.ts.tests.servlet.spec.errorpage.URLClient.servletToDifferentErrorPagesTest 
com.sun.ts.tests.servlet.spec.errorpage.URLClient.wrappedExceptionTest
[servlettck.log](https://github.com/eclipse-ee4j/jakartaee-tck/files/4629760/servlettck.log)

olamy:
@alwin-joseph ah yes I get this. First 2 seems to be wrong, I will fix that (hehe I need to fix that in Jetty as well :) )
Regarding wrappedExceptionTest I need to read specs for wrapped exception, I will look at it today

scottmarlow:
Regarding the 8.0.2 Platform TCK issue, do you have a workaround or will you create a TCK challenge to exclude the offending tests?

The process for challenging TCK tests are mentioned at [TCK Process](https://jakarta.ee/committees/specification/tckprocess/).

olamy:
@scottmarlow you misunderstood my comment (or I wrote it badly :) ).
I agree it's a bit confusing because the TCK until this was simply testing **nothing** because of this attribute name typo. The test was just asserting a servlet request attribute was returning null (but for sure it was because the request attribute name have a typo).
So do we need to go to a complex challenge process (?) whereas it's clearly a bug in the test. 

what I mean in my comment is the 2 tests are failing 
`com.sun.ts.tests.servlet.spec.errorpage.URLClient.heirarchyErrorMatchTest
com.sun.ts.tests.servlet.spec.errorpage.URLClient.servletToDifferentErrorPagesTest`
because my change is wrong for the assert on what should be returned. I looked at Jetty and Jetty is doing the wrong thing here. So I will fix the PR (and fix Jetty as well but it's another issue :)).
The 3rd one I was not sure if the attribute must contain the main exception or the wrapped exception. But after reading the spec it's not really exactly  defined but it would makes more sense as it is used to manage error page mapping. 
So I will change this PR accordingly.
I hope this makes sense?

scottmarlow:
Thanks @olamy, we should focus on fixing the typo and tests on https://github.com/eclipse-ee4j/jakartaee-tck master branch.  Regarding corrections to the test, we should explain why the current test(s), are not Servlet spec compliant and what the fix is, when that is understood.  Make sense?

olamy:
@scottmarlow yes definitely I will close this PR and do the work in another one. Thanks!

src/com/sun/ts/tests/servlet/spec/errorpage/SecondServletErrorPage.java
```diff
@@ -60,7 +61,8 @@ public void service(HttpServletRequest req, HttpServletResponse res)
     pw.println("Servlet Name: " + req.getAttribute(SERVLET_NAME));
     pw.println("Request URI: " + req.getAttribute(REQUEST_URI));
     pw.println("Status Code: " + req.getAttribute(STATUS_CODE));
-    pw.println("Exception Type: " + req.getAttribute(EXCEPTION_TYPE));
+    // with the cast we even enforce it's a Class type
+    pw.println("Exception Type: " + ((Class)req.getAttribute(EXCEPTION_TYPE)).getName());
```
markt-asf, line 65: This triggers a NullPointerException when the request attribute is null causing one TCK test to fail.
I'll include a fix for this in #338.

olamy: is it supposed to be null?

markt-asf: Yes. If an error status code is set without an exception this attribute will be null.

Issue #260, opened 2020-05-13
WebSocket TCK excessively slow and some tests unstable


markt-asf:
Running the standalone WebSocket test suite a significant number of tests take 10s of seconds when, given the actions of the test, they would be expected to take a second or two.

Reviewing the code identified a number of tests that were using a `CountDownLatch` incorrectly to wait for messages to arrive. The calls to `await() ` were timing out rather than the latch being released. The time waiting for the timeout was causing the delay to the tests.

I have a fix and will submit a PR shortly.

Fixing the use of `CountDownLatch` also resolved a long standing issue with a couple of timeout tests where the results were not stable.

scottmarlow:
+100 for this issue, will be nice to have a faster standalone WebSocket test run!  Thanks @markt-asf in advance for the correction!

markt-asf:
See PR #261

markt-asf:
PR applied. This issue has been fixed.

PR #261, opened 2020-05-13
Fix unreliable timeout tests (issue #260)

markt-asf:
I have included a number of commits that just clean-up the code. I'm happy to move them to a separate PR if required.

The priority for me is the commit that fixes the tests. Not only does it reduce TCK runtime from ~30mins to ~20mins (for me, anyway) it fixes the two timeout tests that have been unreliable for as long as I have had access to the TCK.

### DIFF
--- a/src/com/sun/ts/tests/websocket/ee/jakarta/websocket/websocketmessage/WSFullShortAndSessionServer.java
+++ b/src/com/sun/ts/tests/websocket/ee/jakarta/websocket/websocketmessage/WSFullShortAndSessionServer.java
@@ -29,6 +29,7 @@ import com.sun.ts.tests.websocket.common.util.IOUtil;
 @ServerEndpoint("/fullshortsession")
 public class WSFullShortAndSessionServer {
 
+  @SuppressWarnings("unused")
   @OnMessage
   public String echo(Short s, Session session) {
     return String.valueOf(s.shortValue());
--- a/src/com/sun/ts/tests/websocket/api/jakarta/websocket/clientendpointconfig/WSClient.java
+++ b/src/com/sun/ts/tests/websocket/api/jakarta/websocket/clientendpointconfig/WSClient.java
@@ -1150,9 +1150,11 @@ class TCKConfigurator extends Configurator {
   void TCKConfigurator() {
   }
 
+  @Override
   public void beforeRequest(Map<String, List<String>> headers) {
   }
 
+  @Override
   public void afterResponse(HandshakeResponse hr) {
   }
 }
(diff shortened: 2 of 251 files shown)


scottmarlow:
IMO, we should apply bug fixes after we can compile the Platform TCK again.

markt-asf:
Ah. Didn't realise there was a broader compilation issue. I've been compiling individual TCKs (EL and WebSocket so far) and that has been working. I'd be happy to hold off on the clean-up as it is just cosmetic but given the standalone TCK compiles both before and after the bug fix and it only touches WebSocket sources I'd argue that it is safe to merge the bugfix commit from this PR. Let me know if you want me to update / resubmit so you have a PR with just the bug fix.

src/com/sun/ts/tests/websocket/ee/jakarta/websocket/websocketmessage/WSFullShortAndSessionServer.java
```diff
@@ -30,7 +30,7 @@
 public class WSFullShortAndSessionServer {
 
   @OnMessage
-  public String echo(Short s, Session session) {
```
jansupol, line 33: I know the session is not used. But the test is about the method with arguments Short and Session being used. See the class name: `WSFullShortAndSessionServer`. There is another endpoint `WSFullShortServer.java` that has ``echo(Short s)`. So now you have effectively duplicated a test and removed a test that verifies that a method with `Session` argument is invoked at all.

markt-asf: Thanks. I thought I caught all of those but I must have missed that one. I'll update that part of the clean-up and re-check the rest..
Note: I'm still happy to move the clean-up to a separate PR. It is the fix for the slow/unstable tests that is important.

jansupol: It looks like this is removed by mistake, all other similar cases are unchanged

jansupol: Yep, good work on all the tests with CountDownLatch(too_many).

markt-asf:
I've re-based the PR (not that it really needed it but since I was going to have to force push anyway...) and fixed the issue found by Jan. I've also removed some trailing white space I had unintentionally added.

scottmarlow:
Thanks @markt-asf for the contribution!

Thanks @jansupol for reviewing!

PR #263, opened 2020-05-13
Fix #262: Exclude two tests associated with WebSocket spec bug

### LINKED ISSUE #262: Exclude 2 WebSocket tests
There is an [open bug against WebSocket](https://github.com/eclipse-ee4j/websocket-api/issues/228) that can't be fixed until Jakarta EE 10. The short version is that the return value for `Session.getRequestURI()` is not clearly defined in the specification. The TCK currently tests for a return value that is inconsistent with the consensus view of the correct return value. Therefore, the two tests that check this value need to be excluded until the specification is clarified and the TCK updated.

I'll provide a PR shortly.

markt-asf:
The WebSocket issue is:
https://github.com/eclipse-ee4j/websocket-api/issues/228

### DIFF
--- a/install/websocket/bin/ts.jtx
+++ b/install/websocket/bin/ts.jtx
@@ -21,3 +21,6 @@
 # Web Socket 1.0 TCK Exclude List
 #
 
+# Issue https://github.com/eclipse-ee4j/websocket-api/issues/228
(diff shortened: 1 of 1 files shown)


scottmarlow:
Thanks for the change @markt-asf and including a link to the WebSocket tracker.

PR #273, opened 2020-05-14
fix remaining javax.servlet in servlet TCK part (actually in servlet ee)

olamy:
Signed-off-by: olivier lamy <olamy@apache.org>
issue #223


scottmarlow:
Thanks @olamy for the change!  Thanks @gurunrao for the review!

PR #277, opened 2020-05-19
Issue #254 fix typo to display correct exception type/class

olamy:
Currently there is a typo here https://github.com/eclipse-ee4j/jakartaee-tck/blob/a50bf877a5413b812fad569d116c38160ba0aa41/src/com/sun/ts/tests/servlet/spec/errorpage/SecondServletErrorPage.java#L40 
Please note the attribute name `javax.serlvet.error.exception_type` rather than `javax.servlet.error.exception_type`
I guess the test was to assert as defined in the spec the servlet request attribute should contains the class of the exception in the type `java.lang.Class` but this test never fail despite the typo in  the attribute because even the assert is wrong because it's asserting it contains `null` https://github.com/eclipse-ee4j/jakartaee-tck/blob/a50bf877a5413b812fad569d116c38160ba0aa41/src/com/sun/ts/tests/servlet/spec/errorpage/URLClient.java#L149  which will be always true as the servlet request attribute name with the typo is unknown.
The current fix ensure there is no typo but using the constants from servlet-api.jar class RequestDispatcher https://github.com/eclipse-ee4j/jakartaee-tck/pull/277/files#diff-b13cef9ddaf95aab6349e782af97d75eR41 furthermore I even added a cast to `Class` to ensure the servlet request contains the correct attribute https://github.com/eclipse-ee4j/jakartaee-tck/pull/277/files#diff-16e0845578ace345bd725409b0fe1172R64

### DIFF
--- a/src/com/sun/ts/tests/servlet/spec/errorpage/SecondServletErrorPage.java
+++ b/src/com/sun/ts/tests/servlet/spec/errorpage/SecondServletErrorPage.java
@@ -22,6 +22,7 @@ package com.sun.ts.tests.servlet.spec.errorpage;
 
 import com.sun.ts.tests.servlet.common.util.Data;
 
+import jakarta.servlet.RequestDispatcher;
 import jakarta.servlet.ServletException;
 import jakarta.servlet.http.HttpServlet;
 import jakarta.servlet.http.HttpServletRequest;
@@ -35,17 +36,17 @@ import java.io.PrintWriter;
 
 public class SecondServletErrorPage extends HttpServlet {
 
-  private static final String STATUS_CODE = "jakarta.servlet.error.status_code";
+  private static final String STATUS_CODE = RequestDispatcher.ERROR_STATUS_CODE; // "jakarta.servlet.error.status_code";
 
-  private static final String EXCEPTION_TYPE = "jakarta.servlet.error.exception_type";
+  private static final String EXCEPTION_TYPE = RequestDispatcher.ERROR_EXCEPTION_TYPE; //"jakarta.servlet.error.exception_type";
 
-  private static final String MESSAGE = "jakarta.servlet.error.message";
+  private static final String MESSAGE = RequestDispatcher.ERROR_MESSAGE; //"jakarta.servlet.error.message";
 
-  private static final String EXCEPTION = "jakarta.servlet.error.exception";
+  private static final String EXCEPTION = RequestDispatcher.ERROR_EXCEPTION;  //"jakarta.servlet.error.exception"
 
-  private static final String REQUEST_URI = "jakarta.servlet.error.request_uri";
+  private static final String REQUEST_URI = RequestDispatcher.ERROR_REQUEST_URI; //"jakarta.servlet.error.request_uri";
 
-  private static final String SERVLET_NAME = "jakarta.servlet.error.servlet_name";
+  private static final String SERVLET_NAME = RequestDispatcher.ERROR_SERVLET_NAME; //"jakarta.servlet.error.servlet_name"
 
   private static final String EXP_MESSAGE = "error page invoked";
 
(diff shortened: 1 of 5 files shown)


olamy:
fix for #254

src/com/sun/ts/tests/servlet/spec/errorpage/SecondServletErrorPage.java
```diff
@@ -60,7 +61,8 @@ public void service(HttpServletRequest req, HttpServletResponse res)
     pw.println("Servlet Name: " + req.getAttribute(SERVLET_NAME));
     pw.println("Request URI: " + req.getAttribute(REQUEST_URI));
     pw.println("Status Code: " + req.getAttribute(STATUS_CODE));
-    pw.println("Exception Type: " + req.getAttribute(EXCEPTION_TYPE));
+    // with the cast we even enforce it's a Class type
```
scottmarlow, line 64: So, your adding an additional test by casting to Class, to enforce that type Class is returned for STATUS_CODE.  

URLClient.java contains  @assertion_ids for which Servlet spec sections are tested.  Please add an assertion for this additional requirement that your now causing to be enforced.  For example, Servlet:SPEC:104.3.2 is specified in internal/docs/servlet/ServletSpecAssertions.html: 
```The request attribute 'javax.servlet.error.exception_type' of type 'java.lang.Class' will be set if the error page was invoked due to an uncaught exception.
```
internal/docs/servlet:/ServletSpecAssertions.xml:
```
 <assertion required="true" impl-spec="false" defined-by="technology" status="active" testable="true">
                            <id>Servlet:SPEC:104.3.2</id>
                            <description>The request attribute 'javax.servlet.error.exception_type' of type 'java.lang.Class' will be set if the error page was invoked due to an uncaught exception.</description>
                            <location chapter="10" section="9.1"/>
                        </assertion>
```
We also should update the 'javax.servlet' to instead be 'jakarta.servlet' in internal/docs.

Do you agree that the assertion that I pasted already covers the additional test that you are adding with this change?

@markt-asf @stuartwdouglas do you approve this change to enforce that Servlet containers request attribute 'jakarta.servlet.error.exception_type' will be of type java.lang.Class.

src/com/sun/ts/tests/servlet/spec/errorpage/URLClient.java
```diff
@@ -146,7 +146,7 @@ public void wrappedExceptionTest() throws Fault {
     TEST_PROPS.setProperty(SEARCH_STRING, "Second ErrorPage|"
         + "Servlet Name: WrappedException|"
         + "Request URI: /servlet_spec_errorpage_web/WrappedException|Status Code: 500|"
-        + "Exception Type: null|"
```
scottmarlow, line 149: Why are we removing null?

scottmarlow: I think that I understand now, the null was incorrectly checked for (due to the attribute name typo that is also being fixed by this pr).

src/com/sun/ts/tests/servlet/spec/errorpage/SecondServletErrorPage.java
markt-asf: I agree the assertion you pasted already covers this.

The attribute is already defined as Class in the specification document. See section 10.9.1, table 10-1 (servlet 4.0) so I have no objection to the TCK enforcing this.

Overall the PR looks good to me.

olamy:
@scottmarlow any idea if it will be merged soon?

scottmarlow:
@olamy [jakartaee-tck/pull/324](https://github.com/eclipse-ee4j/jakartaee-tck/pull/324) corrected the typo but not the other changes you made (e.g. still need to correct the assertion on null).  

We can merge soon after you update this pr.

olamy:
@scottmarlow rebase done

olamy:
@scottmarlow you can merge it?

olamy:
maybe @gregw and @markt-asf can approve this as well? thanks

olamy:
I wonder what is still expected to get this merged?

PR #288, opened 2020-05-22
remove EJB interop tests.

gurunrao:
Signed-off-by: gurunandan <gurunandan.rao@oracle.com>


install/jakartaee/bin/ts.jte
```diff
@@ -1287,7 +1287,7 @@ command.testExecuteAppClient2= \
         PATH="${javaee.home.ri}/nativelib" \
         APPCPATH=${ts.home}/lib/tsharness.jar${pathsep}${ts.home}/lib/cts.jar${pathsep}${ts.home}/lib/riinterceptors.jar \
 	TZ=${tz} \
-        ${RI_JAVA_HOME}/bin/java \
```
alwin-joseph, line 1290: Just to confirm, since interop tests is now removed we wont need the RI_JAVA_HOME correct ?

scottmarlow: I think that jws + jaxws test folders also use forward/reverse testing with the RI.

gurunrao: RI_JAVA_HOME was  used for EJB interop tests, jdk installation separation.

alwin-joseph: @gurunrao Can you suggest if we can retain the RI_JAVA_HOME, default it to JAVA_HOME so the jaxws tests can make use of it, if required.

gurunrao: have add back RI_JAVA_HOME, so that in future jaxws tests can make used of them.

scottmarlow:
@gurunrao do the interop references in release/tools/build.xml need to be removed? I assume that would include jakartaee-interop + jakartaee-interop-security.

gurunrao:
> @gurunrao do the interop references in release/tools/build.xml need to be removed? I assume that would include jakartaee-interop + jakartaee-interop-security.

@scott have removed jakartaee-interop + jakartaee-interop-security from release/tools/build.xml

scottmarlow:
Thanks Guru!

PR #295, opened 2020-05-26
Fix #294 - standalone JSP TCK failures

### LINKED ISSUE #294: Standalone JSP TCK fails signature tests on Java 11
There are multiple issues here:

- Not using the Jdk9SigTestDriver on Java 11
- When Jdk9SigTestDriver unnecessary packages in the signature files trigger different failures

I have fixes for all of this locally. I just want to run some final checks and then I'll provide a PR.

markt-asf:
Fixes issues with the signature test for the JSp TCK when running on Java 11 in standalone mode.


src/com/sun/ts/tests/signaturetest/signature-repository/jakarta.servlet.jsp.sig_3.0_se8
```diff
@@ -1,425 +1,10 @@
 #Signature file v4.3
-#Version 2.3_se8
+#Version 3.0_se8
```
alwin-joseph, line 2: Hi Mark - This sig file has issues and is being regenerated in https://github.com/eclipse-ee4j/jakartaee-tck/pull/301 to fix . Can you please delete this file.

install/jsp/bin/sig-test-pkg-list_se8.txt
```diff
@@ -22,15 +23,6 @@
...
 ##
 
-java.io
-java.lang
-java.util
-jakarta.el
-jakarta.servlet
 jakarta.servlet.jsp
 jakarta.servlet.jsp.el
-jakarta.servlet.jsp.jstl.core
```
alwin-joseph, line 32: Is it required to remove these. We add the corresponding jars of these packages in the sigTestClasspath while generating the signature file and also to run the tests.

src/com/sun/ts/tests/signaturetest/SigTest.java
```diff
@@ -51,8 +52,12 @@
   protected SignatureTestDriver getSigTestDriver() {
 
     if (driver == null) {
-      driver = SignatureTestDriverFactory
-          .getInstance(SignatureTestDriverFactory.SIG_TEST);
+      String version = (String) System.getProperties().get("java.version");
```
alwin-joseph, line 55: Thanks. Hope you were able to test this change and run sig tests in jdk11.

markt-asf: Yes. I ran the tests with both Java 8 and Java 11 with this PR in place.

install/jsp/bin/sig-test-pkg-list_se8.txt
markt-asf: As far as I could tell, yes. The Java 11 signature test machinery doesn't provide an easy way to exclude packages from the signature tests so you need to make sure they aren't present. There might be a better way of achieving the same result but I couldn't find one.

src/com/sun/ts/tests/signaturetest/signature-repository/jakarta.servlet.jsp.sig_3.0_se8
markt-asf: I'm assuming you mean from the PR. Will do.

markt-asf:
That was fun. I should have removed the changes to the se8 signature files from this PR now. Let me know if you need anything else from me.

install/jsp/bin/sig-test-pkg-list_se8.txt
alwin-joseph: For java 11,  the corresponding file is install/jsp/bin/sig-test-pkg-list_se11.txt. Can you make these changes in this file.

alwin-joseph: I was thinking since sig-test-pkg-list_se8.txt is used for jdk8 runs which should work , was not sure why this is changed. But the packages removed are not required to be tested for jsp standalone I presume.

markt-asf: Correct. Those packages aren't part of the JSP spec. It did work for Java 8 both with and with out those changes. I was trying to avoid future confusion if someone compared the Java 8 signatures with those for Java 11 and wondered why they were wildly different.

Issue #302, opened 2020-06-03
Convert javax.xml.bind=jakarta.xml.bind


scottmarlow:
We have a few references to javax.xml.bind still, that should be renamed in a few tests.  The same tests should be marked as optional (as per https://eclipse-ee4j.github.io/jakartaee-platform/jakartaee9/JakartaEE9ReleasePlan, see "Jakarta XML Binding (Optional)").

scottmarlow:
@alwin-joseph note that https://repo1.maven.org/maven2/javax/enterprise/deploy/javax.enterprise.deploy-api/1.7/javax.enterprise.deploy-api-1.7.jar doesn't have the jakarta.xml.bind interfaces.

scottmarlow:
keyword.properties updated to make the two tests that use jaxb optional:
```
com/sun/ts/tests/webservices12/ejb/annotations/WSEjbWebServiceProviderTest= webservices javaee_optional jaxws_web_profile javaee_web_profile_optional
```

alwin-joseph:
I was hoping the pending occurences of javax* packages would appear in the build logs. But javax.xml.bind in src/com/sun/ts/tests/webservices12/ejb/annotations/WSEjbWebServiceProviderTest/HelloBean.java and others did not come up in the build logs. Thanks for noticing.

scottmarlow:
It is hard to notice, as the javax.xml.bind classes are in JDK 8, so no failures appeared.  I got lucky and noticed it for something else I was working on.

PR #303, opened 2020-06-03
Convert javax.xml.bind=jakarta.xml.bind

scottmarlow:
https://github.com/eclipse-ee4j/jakartaee-tck/issues/302


install/jakartaee/bin/sig-test.map
```diff
@@ -155,7 +155,7 @@ javax.enterprise.deploy=1.2_se6
 jakarta.security.jacc=1.2_se6
 
 # JAXB 
-javax.xml.bind=2.1_se6
+jakarta.xml.bind=3.0_se6
```
alwin-joseph, line 158: This is old sig map file. We use install/jakartaee/bin/sig-test_se8.map for jdk8 runs where we do not have sig test for jakarta.xml.bind mapping or the sig file in this repo. Will need to check whether we need to add jakarta.xml.bind along with platform run. jakarta.xml.bind standalone sig test is running separately though.

scottmarlow: I'll revert this file change, now that I understand that this file is for historical reference (I assume, due to standard edition 6 reference).

scottmarlow: > Copyright year may need to be updated from "Copyright (c) 2018" to "Copyright (c) 2018, 2020"

@RohitKumarJain thanks, I reverted the docker/jaxrtck.sh file change which was the only one missing a copyright date change.  We don't really need to update docker/jaxrtck.sh as we are removing JAXR.

PR #323, opened 2020-06-13
[issues/266] more XSD version corrections

scottmarlow:


scottmarlow:
This should cover the samples tests, still need to find other places where the version needs to be updated.

PR #337, opened 2020-06-17
320 fix xmlns (JCP->Jakarta) in Batch Job XML

### LINKED ISSUE #320: Convert batch (jbatch) TCK test job XMLs from JCP -> Jakarta targetNamespace
Continuing #162 , we still need to update 

`xmlns="http://xmlns.jcp.org/xml/ns/javaee"`
  => 
`xmlns="https://jakarta.ee/xml/ns/jakartaee"`

for the batch job XMLs, e.g.
 **src/com/ibm/jbatch/tck/testJobXml/META-INF/batch-jobs/testMetricCount.xml**

scottkurz:
Most of the fixes are XML files.. there are two Java files that referenced the NS changed too.

Since the sed command I was using was changing the EOL chars I just changed them all in a first commit and did the xmlns change in the second:

```
$ find ...  | xargs sed -i 's%http://xmlns\.jcp\.org/xml/ns/javaee%https://jakarta.ee/xml/ns/jakartaee%'
```

Note another option would have been to simplify replace with the https://github.com/eclipse-ee4j/batch-tck contents, which also have formatting changes.   I think that would be valid, but not as trivially easy to verify, so we can hold off on that until a later date.

Fixes #320


scottmarlow:
@scottkurz what is the 1.0 version number for exactly in ```<job id="chunkSkipInitialTest" xmlns="https://jakarta.ee/xml/ns/jakartaee" version="1.0">```?  Does the version number need to increase to reflect something in EE 9, like https://jakarta.ee/xml/ns/jakartaee/batchXML_2_0.xsd or https://jakarta.ee/xml/ns/jakartaee/jobXML_2_0.xsd?

scottkurz:
Thanks @scottmarlow for bringing that up.. it hadn't crossed my mind.    

The 1.0 matches the API's XSD currently, (so can't just be changed in the TCK), but I see in https://jakarta.ee/xml/ns/jakartaee/ the pattern is to match the spec version.

I opened:  https://github.com/eclipse-ee4j/batch-api/issues/34  to track and will let the platform ML know about this.

If you want to merge I can do another PR later... or we'll have to re-release the API.

scottmarlow:
Thanks @scottkurz, following up with another PR sounds good.  Will merge this pr now.

PR #338, opened 2020-06-17
Servlet 5 fixes

markt-asf:
Still running the TCK. Providing this draft PR to provide visibility of issues (and potential fixes) as I discover therm.


scottmarlow:
I'll review again after/if further changes are made :)

markt-asf:
Just a quick update. This is now passing (with one expected failure because Tomcat always overrides the context path setting in web.xml) for Tomcat 10 on Java 8. There are a handful of failures when using Java 11 (but the signatures do pass). I'm still working on debugging those failures.

markt-asf:
Java 11 issues fixed now.

alwin-joseph:
Just checking, I could see only one failing test (com/sun/ts/tests/servlet/spec/errorpage/URLClient.java#statusCodeErrorPageTest) for jdk8 run with glassfish6. 

exception: 
[javatest.batch] 07-05-2020 11:42:14:  ERROR: [WebValidatorBase] Unexpected Status Code recieved from server.  Expected '501' received '500'

markt-asf:
Some of the failures I saw with Tomcat you might not see with Glassfish (e.g. the CRLF issue).
I'd expect the major version test to fail unless Glassfish hasn't updated its return value for that which is possible.
A number of failures were related to Java 11.
In other words, I am not aware of anything in this PR that shouldn't/doesn't need to be applied but I can build the latest Servlet TCK locally and re-run the tests with the PR applied if that would help.

alwin-joseph:
> Some of the failures I saw with Tomcat you might not see with Glassfish (e.g. the CRLF issue).
> I'd expect the major version test to fail unless Glassfish hasn't updated its return value for that which is possible.
> A number of failures were related to Java 11.
> In other words, I am not aware of anything in this PR that shouldn't/doesn't need to be applied but I can build the latest Servlet TCK locally and re-run the tests with the PR applied if that would help.

I wanted to ask if this change includes fix for the above failing test.  Did not see any issue with this PR. Thanks.

scottmarlow:
@markt-asf could you please signoff on the last two commits that you added?

markt-asf:
Odd. I had signed off on those. I've re-based and forced push and now only one of them is showing as not having the footer. Very strange. I'll double check the sign-off text but it should be generated by my IDE.

markt-asf:
Aha. One of the earlier commits was missing the sign-off but the tool appears to flag the wrong commit. Should be all sorted now.

install/servlet/bin/ts.jte
```diff
@@ -192,7 +192,7 @@ env.ts_win32.menu=true
 # The sigTestClasspath specify the location of servlet implementation 
 # classes that used for signature test
 ##########################################################################
-sigTestClasspath=${java.home}/lib/rt.jar${pathsep}${servlet.classes}
```
scottmarlow, line 195: @markt-asf sorry that I missed this before, but I have a question.  Does install/servlet/bin/ts.jte.jdk11 also need a change?  Currently it has:
```
sigTestClasspath=${java.home}/lib/rt.jar${pathsep}${servlet.classes}
```

markt-asf: The approach of having a separate `ts.jte` file for Java 11 post-dates this PR. Strictly you need `${java.home}/lib/rt.jar${pathsep}${servlet.classes}` for Java 8 and `${java.home}/lib/modules${pathsep}${servlet.classes}` for Java 11. The new version I provided in this PR enables you to use the same file for both Java 8 and Java 11.

Personally, I'd prefer to avoid the use of a Java version specific `ts.jte` file but that isn't my call to make. Happy to adjust this PR however you see fit.

scottmarlow: @markt-asf for some ts.jte files it is easier to have two separate files.

Lets merge this change as is and figure out separately if its possible to have one ts.jte for Servlet versus two.

scottmarlow: CC @alwin-joseph ^

Issue #339, opened 2020-06-17
Add code back in to create the ts_dep folder in the dist folder


scottmarlow:
Thanks @gerdogdu for finding this bug and posting to the jakartaee-tck-dev@eclipse.org mailing list, very much appreciated!

https://gist.github.com/scottmarlow/67a84984df5d449560d6691f1c7b2b22 shows select output from ```git diff --no-ext-diff --unified=0 8.0.2 HEAD --no-ext-diff > allchanges.diff``` that was removed with other Jakarta Deployment handling code.  The problem is that the removed getClientClassPath() code had a side effect of creating the ts_dep folder in the dist.

scottmarlow:
We probably need add the folder creation code just under line 128 in com.sun.ts.lib.implementation.sun.javaee.glassfish.AutoDeployment#init:

```
      TestUtil.logHarness(
          "Creation of TSDeployment implementation instance failed."
              + " Please check the values of 'porting.ts.deploy2.class.1' and"
              + " 'porting.ts.deploy2.class.2'");
    } // line # 128
```
Something like:
```
    }
    if (javaeeLevel.contains("full")) {
       Insert ts_dir create folder code
    }
```

PR #342, opened 2020-06-19
Fix Client Stub Generation for App Clients

gerdogdu:
This code fixed client stub generation via calling asaadmin get-client-stubs from -deploy target in deploy.xml. Please also see, https://github.com/eclipse-ee4j/jakartaee-tck/issues/339


src/com/sun/ts/lib/implementation/sun/javaee/glassfish/AutoDeployment.java
```diff
@@ -197,6 +197,26 @@ public String deploy(DeploymentInfo info) throws TSDeploymentException {
     Project p = new Project();
     antSetupFromInfo(info, p);
 
+    String javaeeLevel = propMgr.getProperty("javaee.level", "full");
```
scottmarlow, line 200: @gerdogdu I'm not sure if it is required but from an ordering perspective, it may be better to move this new code after the current call to p.executeTarget("-deploy").  Perhaps in place of where getClientClassPath was called from.  The advantage is that we would then be sure that application archive deploys before the client stubs archive is generated.

scottmarlow: The change looks correct otherwise.  I think we should merge it after you update the ordering.  Thanks for investigating and contributing the fix!  Very much appreciated! :+1:

gerdogdu:
@scottmarlow Actually it must be before the calling target -deploy in deploy.xml because **get.stub.clients** antcall is called after -deploy target. This block of code before p.executeTarget("-deploy"); only sets the properties which are used by ant targets if it is executed "full" profile. 

- deploy.app.name= App name

- get.stub.clients = If set, then we call the target ant call via  <antcall target="get.client.stubs"> in deploy.xml.

- get.stub.clients.ts_dep= where to put dependencies

If the get.stub.clients is set by the code, then stub generation is executed.
Hope it clears.
Gurkan

gerdogdu:
@scottmarlow Also for your question,
 We sure that application archive deploys before the client stubs archive is generated (please see deploy.xml file, <target name="-deploy">). Stub generations antcalled is executed after deployment.

scottmarlow:
Thanks for explaining @gerdogdu!

bin/xml/impl/glassfish/deploy.xml
```diff
@@ -33,28 +31,40 @@
...
+			<fileset refid="deploy.archive.set" />
+		</ts.add.sun.xml.to.component.archive>
+
+		<ts.glassfish.deploy>
+			<fileset refid="deploy.archive.set"/>
+		</ts.glassfish.deploy>
+		
+		<!-- Client stub generation-->
+		<if>
+			<isset property="get.stub.clients"/>
```
scottmarlow, line 59: @gerdogdu should we undeploy the EJB client stubs when the client application is undeployed?

gerdogdu:
@scottmarlow Does the old version of TCK removes them? If so, we need to remove then. I think just remove hte ts_dep folder

PR #352, opened 2020-07-01
disable GlassFish 6.0 resource validation for mail tests via create-jvm-options

scottmarlow:

### DIFF
--- a/docker/run_jakartaeetck.sh
+++ b/docker/run_jakartaeetck.sh
@@ -396,7 +396,7 @@ if [[ $test_suite == "javamail" || $test_suite == "samples" || $test_suite == "s
   ESCAPED_MAIL_USER=`echo ${MAIL_USER} | sed -e 's/@/%40/g'`
   cd  ${TS_HOME}/bin
   ant -DdestinationURL="imap://${ESCAPED_MAIL_USER}:${MAIL_PASSWORD}@${MAIL_HOST}:${IMAP_PORT}" populateMailbox
-  export DISABLE_RESOURCE_VALIDATION="-Ddeployment.resource.validation=false"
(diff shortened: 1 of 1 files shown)


docker/run_jakartaeetck.sh
```diff
@@ -409,7 +409,7 @@ ant config.ri
 cd ${CTS_HOME}
 export PORT=5858
 ${CTS_HOME}/ri/${GF_RI_TOPLEVEL_DIR}/glassfish/bin/asadmin --user admin --passwordfile ${ADMIN_PASSWORD_FILE} -p ${PORT} stop-domain
-${CTS_HOME}/ri/${GF_RI_TOPLEVEL_DIR}/glassfish/bin/asadmin $DISABLE_RESOURCE_VALIDATION --user admin --passwordfile ${ADMIN_PASSWORD_FILE} -p ${PORT} start-domain
+${CTS_HOME}/ri/${GF_RI_TOPLEVEL_DIR}/glassfish/bin/asadmin --user admin --passwordfile ${ADMIN_PASSWORD_FILE} -p ${PORT} start-domain
```
smillidge, line 412: This won't set it on GlassFish I will need to take a look at the TCK source and think of how to do this.

scottmarlow: The above line is reverting the previous change.  

Would ```glassfish/bin/asadmin create-jvm-options -Ddeployment.resource.validation=false``` work?  That is how we set other options, like ```-Dosgi.shell.telnet.port=6667```

scottmarlow: https://eclipse-ee4j.github.io/glassfish/docs/5.1.0/administration-guide/jvm.html#ggnes describes the above which is what I used in this change.

smillidge: Yes that would work.

PR #354, opened 2020-07-04
Changes to run standalone TCKs using JDK11

alwin-joseph:
Signed-off-by: alwin-joseph <alwin.joseph@oracle.com>

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -181,6 +181,8 @@ spec:
            description: 'List of standalone TCK bundle file names to be appended to the base url' )
     choice(name: 'PROFILE', choices: 'FULL\nWEB', 
            description: 'Profile to be used for running CTS either web/full' )
+    choice(name: 'JDK', choices: 'JDK8\nJDK11',
+           description: 'Java SE Version to be used for running TCK either JDK8/JDK11' )
     choice(name: 'LICENSE', choices: 'EPL\nEFTL',
            description: 'License file to be used to build the TCK bundle(s) either EPL(default) or Eclipse Foundation TCK License' )
     choice(name: 'DATABASE', choices: 'JavaDB\nOracle\nMySQL', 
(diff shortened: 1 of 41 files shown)


Jenkinsfile
```diff
@@ -181,6 +181,8 @@ spec:
            description: 'List of standalone TCK bundle file names to be appended to the base url' )
     choice(name: 'PROFILE', choices: 'FULL\nWEB', 
            description: 'Profile to be used for running CTS either web/full' )
+    choice(name: 'JDK', choices: 'JDK8\nJDK11',
```
scottmarlow, line 184: Looks good, so I think that the first listed option is taken as a default, so I think we will default to JDK8 if the parameter is not specified.

docker/cajtck.sh
```diff
@@ -55,8 +55,18 @@ echo "TS_HOME $TS_HOME"
 chmod -R 777 $TS_HOME
 cd $TS_HOME/bin
 
+if [[ "$JDK" == "JDK11" || "$JDK" == "jdk11" ]];then
+  export JAVA_HOME=${JDK11_HOME}
+  export PATH=$JAVA_HOME/bin:$PATH
+  cp ts.jte.jdk11 ts.jte
+else
+  sed -i "s#^endorsed.dirs=.*#endorsed.dirs=$TCK_HOME/$GF_TOPLEVEL_DIR/glassfish/modules/endorsed#g" ts.jte
```
scottmarlow, line 63: Looks good, if we get class not found errors with JDK11 that will be a hint that we should see if jars in endorsed folder might need to be on the classpath.  If no errors occur, even better.

Jenkinsfile
alwin-joseph: Yes the default value for variable "JDK" will be JDK8 (the first option) .  Also by default JAVA_HOME will be JDK8 as it is set in the docker image.

docker/cajtck.sh
alwin-joseph: Right.

PR #393, opened 2020-07-30
Also set -Ddeployment.resource.validation=false for EJB30 tests

scottmarlow:

### DIFF
--- a/docker/run_jakartaeetck.sh
+++ b/docker/run_jakartaeetck.sh
@@ -399,7 +399,7 @@ if [[ $test_suite == "javamail" || $test_suite == "samples" || $test_suite == "s
 fi
 ### populateMailbox for javamail suite - End ###
 
-if [[ $test_suite == "javamail" || $test_suite == "samples" || $test_suite == "servlet" || $test_suite == "appclient" || $test_suite == "ejb" || $test_suite == "jsp" ]]; then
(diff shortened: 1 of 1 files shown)


docker/run_jakartaeetck.sh
```diff
@@ -399,7 +399,7 @@ if [[ $test_suite == "javamail" || $test_suite == "samples" || $test_suite == "s
 fi
 ### populateMailbox for javamail suite - End ###
 
-if [[ $test_suite == "javamail" || $test_suite == "samples" || $test_suite == "servlet" || $test_suite == "appclient" || $test_suite == "ejb" || $test_suite == "jsp" ]]; then
+if [[ $test_suite == "javamail" || $test_suite == "samples" || $test_suite == "servlet" || $test_suite == "appclient" || $test_suite == "ejb" || $test_suite == "ejb30" || $test_suite == "jsp" ]]; then
```
alwin-joseph, line 402: I think for ejb30 it should be  [[ $test_suite == ejb30* ]] || [[ "ejb30" == $test_suite ]] as the parameter we pass is like ejb30/lite/tx etc, similar to line 249.

scottmarlow: Excellent point @alwin-joseph, thanks for noticing that!  I switched to ```[[ $test_suite == ejb* ]]``` and followed the same style for the other checks on this line.

PR #398, opened 2020-07-30
show /etc/os-release in output so we know which OS Platform was used for test run

scottmarlow:


scottmarlow:
@alwin-joseph I agree that CentOS 7 is likely used, still I would like to see confirmation.  I also noticed that https://github.com/eclipse-cbi/jiro/blob/master/instances/ee4j.jakartaee-tck/target/jenkins/configuration.yml is referencing both CentOS 7 + CentOS 8 which is why I want to show the additional information.

scottmarlow:
+1, I will move this.

PR #401, opened 2020-07-30
Updating DEFAULT_GF_BUNDLE_URL value and log messages.

RohitKumarJain:


Jenkinsfile
```diff
@@ -203,7 +203,7 @@ spec:
     MAIL_USER="user01@james.local"
     MAIL_HOST="localhost"
     LANG="en_US.UTF-8"
-    DEFAULT_GF_BUNDLE_URL="https://repo1.maven.org/maven2/org/glassfish/main/distributions/glassfish/5.1.0/glassfish-5.1.0.zip"
+    DEFAULT_GF_BUNDLE_URL="https://download.eclipse.org/ee4j/glassfish/glassfish-6.0.0-SNAPSHOT-nightly.zip"
```
scottmarlow, line 206: @RohitKumarJain for the most part, I think that we are always setting the GlassFish zip link in our Jenkins jobs.  So, this is fine for now but perhaps we should update Jenkinsfile to reference https://repo1.maven.org/maven2/org/glassfish/main/distributions/glassfish/6.0.0/glassfish-6.0.0.zip when that becomes available.

PR #446, opened 2020-08-14
Userguide update to remove Apache Ant from tck packages

RohitKumarJain:
Signed-off-by: RohitKumarJain <rohit.ku.jain@oracle.com>


edbratt:
This change will change the TCK which has already been promoted. Please be sure to re-promote and send a notification to the API committer list that this promoted TCK has been updated. (We will need to update the TCK Results documentation as well as any CCR issues)

PR #483, opened 2020-08-27
Restore unaffected javax packages in Assertion html files

dblevins:
A few javax-to-jakarta over corrections that need to be restored back to javax

There are others, but again trying to keep PRs focused so they are more easily mergable.


install/jaspic/docs/assertions/JASPICSpecAssertions.html
```diff
@@ -3429,7 +3429,7 @@ <h2>
             <TD align="center" valign="center"><font size="1PT">3</font></TD>
             <TD align="left" valign="center"><font size="1PT"> This profile requires that the
                     CallbackHandler provided by the ServerAuthModule to the LoginContext constructor support the
-                    jakarta.security.auth.callback.NameCallback and the jakarta.security.auth.callback.PasswordCallback.
+                    javax.security.auth.callback.NameCallback and the javax.security.auth.callback.PasswordCallback.
```
scottmarlow, line 3433: good catch.

PR #520, opened 2020-09-22
Tweak XML WS TCK user guide

lukasj:
* fix xml ws ug toc
* Fix JAXWS acronym
* Change default CI name - use former jaxws-ri instead of full GlassFish
* no need to deal with plugins when running xml ws TCKs


user_guides/jaxws/src/main/jbake/content/attributes.conf
```diff
@@ -1,9 +1,9 @@
 :TechnologyFullName: Jakarta XML Web Services
 :TechnologyShortName: XML Web Services
-:LegacyAcronym: JWB
+:LegacyAcronym: JAXWS
 :TechnologyVersion: 3.0
-:TechnologyRI: Eclipse GlassFish 6.0
-:TechnologyRIURL: https://projects.eclipse.org/projects/ee4j.glassfish
+:TechnologyRI: Eclipse Metro 3.0.0
```
scottmarlow, line 5: Makes sense as long as we pass the TCK with Metro 3.0.0

scottmarlow (approved):

user_guides/jaxws/src/main/jbake/content/attributes.conf
lukasj: If Metro won't pass TCK, GlassFish won't pass it neither, since GlassFish includes Metro. In fact, standalone TCKs were designed for a subset of metro (former jax-ws-ri, currently metro-jax-ws) which used to be part of JDK.

lukasj: ... note that the main difference between Metro and GF is that Metro does not implement Enterprise WS (former JSR-109; TCKs for this are part of the platform TCK), which means that passing TCKs with Metro does not guarantee that GF will pass too.

PR #510, opened 2020-09-16
Build fix for jaxws namespaces

senivam:
Signed-off-by: Maxim Nesen <maxim.nesen@oracle.com>

PR #532, opened 2020-09-30
Revert "Build fix for jaxws namespaces"

alwin-joseph:
Reverts eclipse-ee4j/jakartaee-tck#510


alwin-joseph:
This is to revert the temporary build fix due to schema changes in jaxws ri. The glassfish was updated with the latest jaxws jars and now this change is no longer required.

PR #349, opened 2020-07-01
Pages, Connector and JAXRS TCK fixes

gurunrao:
changes for Pages TCK (fix for two failures), JAXRS TCK(fix for 57 failures), Connector TCK(fix for 138 failures)
CI test run with fix can be found @ https://ci.eclipse.org/jakartaee-tck/job/jakartaee-tck-grao/job/connector-tck/3/

Issue #266, opened 2020-05-14
Update lib/schemas references (e.g. install, release, tests under src) to use either EE 9 or Jakarta EE 8 schemas (as per Jan 21, 2020 VOTE: backward compatibility for descriptor schemas).  Also drop unused lib/schemas + lib/dtds


cesarhernandezgt:

scottmarlow:
Also switch tests to Jakarta EE 9 XSDs + DTDs.

scottmarlow:
I don't see many references to /lib/schemas/*.xsd files, just the following:
```
ag --ignore .git lib.schemas.*xsd
src/com/sun/ts/tests/jaxws/common/xml/catalog/META-INF/jax-ws-catalog.xml
29:      uri="../../../../../../../../../../lib/schemas/ws-addr.xsd"/>
32:      uri="../../../../../../../../../../lib/schemas/swaref.xsd"/>

release/tools/jstl.xml
126:                    <fileset dir="${ts.home}/lib/schemas" includes="web-app*xsd,web-jsptaglibrary*xsd,
```
We can remove the other XSD files from lib/schemas and see if that causes any issues.  This can be done separately I think than the src changes to use the EE9 schemas.

scottmarlow:
Also need to replace 
 http://java.sun.com/xml/ns/j2ee/web-jsptaglibrary_2_0.xsd with http://jakarta.ee/xml/ns/jakartaee/web-jsptaglibrary_3_0.xsd, once https://github.com/eclipse-ee4j/jakartaee-schemas/pull/9 is merged.

scottmarlow:
Remaining references to http://xmlns.jcp.org/xml/ns/javaee:
```
src/com/ibm/jbatch/tck/tests/jslxml/ContextsGetIdTests.java
src/com/ibm/jbatch/tck/testJobXml/META-INF/batch-jobs/nullChkPtInfo.xml
src/com/ibm/jbatch/tck/testJobXml/META-INF/batch-jobs/job_batchlet_step_listener.xml
src/com/ibm/jbatch/tck/testJobXml/META-INF/batch-jobs/startLimitTests.xml
src/com/ibm/jbatch/tck/testJobXml/META-INF/batch-jobs/step_level_properties_count.xml
src/com/ibm/jbatch/tck/testJobXml/META-INF/batch.xml
src/com/ibm/jbatch/tck/testJobXml/META-INF/batch-jobs/decider_as_first_job_element_fails.xml
[...]

PR #536, opened 2020-10-02
Fix #535. Reverts #349 and address root cause of JSP test failure

### LINKED ISSUE #535: com.sun.ts.tests.jsp.spec.el.jsp.ELJspVersionTag relies on implementation detail
com.sun.ts.tests.jsp.spec.el.jsp.ELJspVersionTag was updated for EE9 as part of PR #349 with:

`JspWriter out = getJspContext().getOut();`
`out.println(poundExpr);`

And expects a result of "ValueExpression[#{foo}]".  But this behavior is relying on the RI's [overridden toString method](https://github.com/eclipse-ee4j/el-ri/blob/738dd355879d88f09269a53ab0418d4b5f9ee46d/impl/src/main/java/com/sun/el/ValueExpressionImpl.java#L260).

In other implementations, where this toString method is not overridden, the out.println statement simply results in output like "org.apache.jasper.el.JspValueExpression@20562".  This can be seen in testing with both Tomcat and Open Liberty.

A better solution here would be to use something in the API for validation, like:
`out.println(poundExpr.getExpressionString());`

This would result in the old output of "#{foo}" if everything is working properly.

markt-asf:
The EL JSP version test is testing that TLDs before version 2.1 ignore
the #{...} syntax. Therefore the el_jsp_2.0.tld file needs to use a
schema version less than 2.1 - in this case it uses 2.0.
Also fixes the validity of the el_jsp.tld file.

### DIFF
--- a/src/com/sun/ts/tests/jsp/spec/el/jsp/ELJspVersionTag.java
+++ b/src/com/sun/ts/tests/jsp/spec/el/jsp/ELJspVersionTag.java
@@ -17,8 +17,6 @@
 package com.sun.ts.tests.jsp.spec.el.jsp;
 
 import com.sun.ts.tests.jsp.common.util.JspTestUtil;
-import jakarta.el.ELContext;
-import jakarta.el.ValueExpression;
 import jakarta.servlet.jsp.JspWriter;
 import jakarta.servlet.jsp.JspException;
 import jakarta.servlet.jsp.tagext.SimpleTagSupport;
@@ -26,15 +24,15 @@ import java.io.IOException;
 
 public class ELJspVersionTag extends SimpleTagSupport {
 
-  private ValueExpression poundExpr;
+  private String poundExpr;
 
-  public void setPoundExpr(ValueExpression poundExpr) {
+  public void setPoundExpr(String poundExpr) {
     this.poundExpr = poundExpr;
   }
 
   public void doTag() throws JspException, IOException {
-    ELContext elContext = getJspContext().getELContext();
     JspWriter out = getJspContext().getOut();
+
     try {
       out.println(poundExpr);
     } catch (Throwable t) {
--- a/src/web/jsp/spec/el/jsp/ELJspVersionTest.gf
+++ b/src/web/jsp/spec/el/jsp/ELJspVersionTest.gf
@@ -1,1 +1,1 @@
-ValueExpression[#{foo}]
+#{foo}
--- a/src/web/jsp/spec/el/jsp/WEB-INF/el_jsp.tld
+++ b/src/web/jsp/spec/el/jsp/WEB-INF/el_jsp.tld
@@ -20,7 +20,7 @@
 <taglib xmlns = "https://jakarta.ee/xml/ns/jakartaee"
     xmlns:xsi = "http://www.w3.org/2001/XMLSchema-instance" 
     xsi:schemaLocation = "https://jakarta.ee/xml/ns/jakartaee https://jakarta.ee/xml/ns/jakartaee/web-jsptaglibrary_3_0.xsd" 
-    version = "2.1">
+    version = "3.0">
 
     <tlib-version>1.0</tlib-version>
     <short-name>el</short-name>
--- a/src/web/jsp/spec/el/jsp/WEB-INF/el_jsp_2.0.tld
+++ b/src/web/jsp/spec/el/jsp/WEB-INF/el_jsp_2.0.tld
@@ -17,10 +17,10 @@
 
 -->
 
-<taglib xmlns="https://jakarta.ee/xml/ns/jakartaee"
+<taglib xmlns="http://java.sun.com/xml/ns/j2ee"
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-    xsi:schemaLocation="https://jakarta.ee/xml/ns/jakartaee https://jakarta.ee/xml/ns/jakartaee/web-jsptaglibrary_3_0.xsd"
-    version="3.0">
+    xsi:schemaLocation="http://java.sun.com/xml/ns/j2ee http://java.sun.com/xml/ns/j2ee/web-jsptaglibrary_2_0.xsd"
+    version="2.0">
 
     <tlib-version>1.0</tlib-version>
     <short-name>el_jsp_version</short-name>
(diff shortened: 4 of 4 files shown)


scottmarlow:
@markt-asf thank you very much for responding to this issue on the mailing list/issue and for creating this pull request to address the concern.

If the purpose of this test is specifically to test compatibility with a Java EE version that is less than Java EE 8, then our choices for the test are:

1. Update to the Jakarta EE 9 schema and change the purpose of the test to fit using the Jakarta EE 9 schema.
2. Remove the test as since the test purpose was to test schemas from before Java EE 8 ([8.6. Jakarta EE Application XML Schema](https://jakarta.ee/specifications/platform/9/platform-spec-9-SNAPSHOT.html#a3203) + [Appendix A: Previous Version Deployment Descriptors](https://jakarta.ee/specifications/platform/9/platform-spec-9-SNAPSHOT.html#a3447)

What do you think we should do?

markt-asf:
I think we should do option 3 - leave the test in place and apply this PR.
While the test was written using a tag class it could have been written with a tag file and - if it had - the package rename would have no impact on it. Therefore, users migrating applications that use 2.0 tag files will expect the behaviour that this test is checking. They certainly won't be expecting to have to rewrite all their tags as version 3.0 tag files. That has the potential to impose a significant extra burden on users during upgrade. It would make it all but impossible for the various automated upgrade tools to upgrade an app that used 2.0 tags.

If there is an objection to doing this on the grounds there is no requirement to support prior schema (I'd argue there is a requirement) then there is option 4. Applying this PR and then add the test to the default excludes so containers that want to test this feature can and those that don't can skip it when obtaining certification.

I have a very strong preference for option 3. I prefer option 4 over option 2 but don't like either 4 or 2 and would object to either of those approaches. Option 1 is pointless as there are already tests that cover that usage with a 3.0 schema.

pnicolucci:
This test looks to have been around for awhile. Why would we need to exclude it now 
if it's always been testing an older version of the the tag files? Is this a new requirement
for Jakarta EE9?

scottmarlow:
> I think we should do option 3 - leave the test in place and apply this PR.
> While the test was written using a tag class it could have been written with a tag file and - if it had - the package rename would have no impact on it. Therefore, users migrating applications that use 2.0 tag files will expect the behaviour that this test is checking. They certainly won't be expecting to have to rewrite all their tags as version 3.0 tag files. That has the potential to impose a significant extra burden on users during upgrade. It would make it all but impossible for the various automated upgrade tools to upgrade an app that used 2.0 tags.

EE 9 implementations could support J2EE 1.4 XSDs but that is not required by EE 9 implementations.  We have been removing such `< EE 8` compatibility tests as we discover them. I started drafting the [Jakarta EE 9 Platform TCK rules](https://docs.google.com/document/d/1rdQCftJHexDouNtrvvUfjknZHhG8sWt1lerwcWpbQrY/edit#heading=h.sygoumfn3qd0) which I believe that we have been consistently following.  

A week ago, I was on the same page as you, in that we were considering using the TCK keyword.properties to make some removed technologies be optional, but it was correctly pointed out that would not be consistent with the way we handled other removed technology tests.  In support of trying to ensure we are consistent about the rules that we follow when removing legacy compatibility tests, I started drafting [the Jakarta EE 9 Platform TCK rules](https://docs.google.com/document/d/1rdQCftJHexDouNtrvvUfjknZHhG8sWt1lerwcWpbQrY/edit#heading=h.sygoumfn3qd0) so we can have something in writing (to be moved to the Platform TCK wiki or github pages).

> 
> If there is an objection to doing this on the grounds there is no requirement to support prior schema (I'd argue there is a requirement) then there is option 4. Applying this PR and then add the test to the default excludes so containers that want to test this feature can and those that don't can skip it when obtaining certification.
> 
> I have a very strong preference for option 3. I prefer option 4 over option 2 but don't like either 4 or 2 and would object to either of those approaches. Option 1 is pointless as there are already tests that cover that usage with a 3.0 schema.

If we are going to remove this test, I would like to merge this pr first and tag the commit that removes the commit so we could bring it back into the legacy TCK test kit if we ever create that for legacy tests.

scottmarlow:
> This test looks to have been around for awhile. Why would we need to exclude it now
> if it's always been testing an older version of the the tag files? Is this a new requirement
> for Jakarta EE9?

We have removed legacy (pre EE 8) tests as we discover them, as EE 9 server implementations are not required to support them.  We would of liked to have created a separate legacy TCK test kit to move such tests to but do not likely have time for that (would need to be discussed and process agreed on).

gurunrao:
+1 for merging the PR and removal of test posts post PR merge.

gurunrao (approved):
PR can be merged and tests can be removed post merge of PR.

RohitKumarJain (commented):
We may need to update Copyright year, from Copyright (c) 2018 to Copyright (c) 2018, 2020 in below file:
src/web/jsp/spec/el/jsp/WEB-INF/el_jsp_2.0.tld

markt-asf:
No need to update copyrights. The files will end up unchanged from their original content.
Regarding pre-EE 8 schemas, the only discussions I recall (all in projects) concluded that the correct thing to do was to retain support for those schemas.
I can't access [Jakarta EE 9 Platform TCK](https://docs.google.com/document/d/1rdQCftJHexDouNtrvvUfjknZHhG8sWt1lerwcWpbQrY/edit#heading=h.sygoumfn3qd0) rules so I can't comment on the content but from the context I suspect I'll object strongly to its contents.

scottmarlow:
> Regarding pre-EE 8 schemas, the only discussions I recall (all in projects) concluded that the correct thing to do was to retain support for those schemas.

If you disagree with my assessment that EE 9 implementations can still support handling of the legacy (pre-EE 8 schemas) but the Platform TCK should only test EE 8+ schemas as per https://github.com/eclipse-ee4j/jakartaee-tck/issues/266, that can best be discussed on a mailing list (perhaps Jakarta EE platform mailing list would be appropriate).  

> I can't access [Jakarta EE 9 Platform TCK](https://docs.google.com/document/d/1rdQCftJHexDouNtrvvUfjknZHhG8sWt1lerwcWpbQrY/edit#heading=h.sygoumfn3qd0) rules so I can't comment on the content but from the context I suspect I'll object strongly to its contents.

Sorry, I just shared with all as read only.  I also shared specifically with you to give you comment ability, if you want to comment directly on it.  I also mentioned it yesterday on the Platform TCK mailing list under `Removal of tests for pruned technologies versus making tests optional and not run by default..` thread, which is a more visible place to give feedback, as more people will see it.

scottmarlow:
Already merged via https://github.com/eclipse-ee4j/jakartaee-tck/pull/545

PR #544, opened 2020-10-13
Update copyright years to 2020

dblevins:
Bulk update of all copyright years.

We still have a very large amount of work to do getting RMI-IIOP references out of the TCK with around 300+ references in java files of tests we'll still need.  We'll need to either ignore copyright years for the short term or do a bulk replace now.  It won't scale to do them one at a time by hand on each PR.


internal/docs/jaspic/index.html
```diff
@@ -121,7 +121,7 @@ <h2>JT Harness Documentation</h2>
 
     <hr>
     <p>
-        <cite><small><address>Copyright (c) 2007, 2028 Oracle and/or its affiliates. All rights
```
dblevins, line 124: Interesting a 2028 date slipped in.

arjantijms:
> It won't scale to do them one at a time by hand on each PR.

In Eclipse you can do a search and replace on a targeted resource selection, and that selection can be the files staged for a commit. That makes it relatively easy to do a date replacement in a PR.

bin/xml/impl/glassfish/jersey.xml
```diff
@@ -1,6 +1,6 @@
 <!--
 
-    Copyright (c) 2018 Oracle and/or its affiliates. All rights reserved.
+    Copyright (c) 2018, 2020 Oracle and/or its affiliates. All rights reserved.
```
alwin-joseph, line 3: Is the copyright update policy to update even the untouched files with 2020 ?
In my understanding only the files which were modified recently need to have these updates.

scottmarlow:
Added `future` label to reflect targeting EE 9.1 or whatever the version after EE 9 is.

dblevins:
Closing this PR as if we aren't going to merge it it's just going to collect merge conflicts.  It will be easier to just create a new PR.

That said we need to have a conversation about copyright.  If we are going to be strict and only require/allow the copyrights updated files to be change we should also be strict on what company is attributed in the updated copyright.  In open source a file may start out with a single copyright owners, but as it is edited it legally gains additional copyright owners -- they legally own the copyright of the changes in their diff.  Once merged, the file now has two copyright owners.  There is no legal framework that transfers copyright of the new work to the first-creator of the file.  Therefore, if we are going to be strict and to the letter, a change to a file made by someone who is not employed by Oracle requires much more than an update to the year.

PR #550, opened 2020-10-20
[548] Remove RMI-IIOP tests via keyword.properties change

scottmarlow:
Signed-off-by: Scott Marlow <smarlow@redhat.com>

Addresses https://github.com/eclipse-ee4j/jakartaee-tck/issues/548 in the most minimal way.  Will also create a pull request for removing the tests.

Issue #548, opened 2020-10-20
Remove RMI-IIOP tests


scottmarlow:
https://github.com/eclipse-ee4j/jakartaee-tck/issues/508 made the RMI-IIOP tests optional but that turned out to be wrong, as we do not need to test the Corba RMI-IIOP protocol which is what the RMI-IIOP tests were doing.  

Description of the tests to be removed:
The RMI-IIOP tests were covering various aspects of Corba testing to ensure that the underlying Corba implementation could also handle EJB remote invocations (testing marshalling/unmarshalling, passing complex object graphs that contained several references to a common object, calls that return certain rmi-iiop types and calls between different containers).

scottmarlow:
https://github.com/scottmarlow/jakartaee-tck/tree/removeiiop contains the deep removal of the RMI-IIOP tests, will push a safer version of the change also.

scottmarlow:
https://github.com/scottmarlow/jakartaee-tck/tree/removeiiop_safer only updates keyword.properties to remove the rmi-iiop tests

PR #551, opened 2020-10-20
[548] Remove RMI-IIOP tests

scottmarlow:
Addresses https://github.com/eclipse-ee4j/jakartaee-tck/issues/548 by completely removing the rmiiiop tests.  

If we merge this change, we don't need the safer https://github.com/eclipse-ee4j/jakartaee-tck/pull/550.

If we merge https://github.com/eclipse-ee4j/jakartaee-tck/pull/550, we could consider postponing [this change](https://github.com/eclipse-ee4j/jakartaee-tck/pull/551) until after EE 9.


RohitKumarJain (approved):

gurunrao:
nice to clean-up rmiiiop disable logic here (since post this PR it will no longer be required)- https://github.com/eclipse-ee4j/jakartaee-tck/blob/master/bin/xml/ts.top.import.xml#L429

scottmarlow:
The initial test run that requested testing for `signaturetest samples rmiiiop` worked as expected.  https://ci.eclipse.org/jakartaee-tck/job/jakartaee-tck-scottmarlow/job/removeiiop/4/testReport/ shows that the rmiiiop tests were not run but the signaturetest+samples were.

scottmarlow:
> nice to clean-up rmiiiop disable logic here (since post this PR it will no longer be required)- https://github.com/eclipse-ee4j/jakartaee-tck/blob/master/bin/xml/ts.top.import.xml#L429

:+1:  for including that in your https://github.com/eclipse-ee4j/jakartaee-tck/compare/master...gurunrao:revert-corba-changes?expand=1 change.

scottmarlow:
Since this pull request is working better than https://github.com/eclipse-ee4j/jakartaee-tck/pull/550, I will start a Full Platform TCK run soon that verifies that no failures are introduced by this change.

jeanouii:
If this one works fine, we can probably do some more cleanup. 
I can still see some IIOP mentions mainly GF configuration.
Some other in assertions that we need to look at on a case by case approach I believe.

Good job though

scottmarlow:
> If this one works fine, we can probably do some more cleanup.

Last nights testing showed it works fine.

> I can still see some IIOP mentions mainly GF configuration.

I will look for IIOP mentions.

> Some other in assertions that we need to look at on a case by case approach I believe.

I will look for other assertions that mention IIOP.

> 
> Good job though

gurunrao (approved):

scottmarlow:
> > If this one works fine, we can probably do some more cleanup.
> 
> Last nights testing showed it works fine.
> 
> > I can still see some IIOP mentions mainly GF configuration.
> 
> I will look for IIOP mentions.

I started removing some of these and will create a separate pull request for them, along with the assertion changes.

> 
> > Some other in assertions that we need to look at on a case by case approach I believe.

Thanks for pointing this out.  I am thinking of making these further changes as a separate pull request that doesn't have to be merged today.  The assertion updates may involve discussion that could delay change from being merged by tomorrow, which IMO could impact our schedule.

> 
> > Good job though

Thanks :-)

user_guides/jakartaee/src/main/jbake/content/config.adoc
```diff
@@ -1041,7 +1003,7 @@ ant populateMailbox
 
 [[GJKPO]][[jakarta-restful-web-services-test-setup]]
 
-5.4.9 Jakarta RESTful Web Services Test Setup
+5.4.8 Jakarta RESTful Web Services Test Setup
```
alwin-joseph, line 1006: Might need to correct the following 5.4.9.1 to 5.4.8.1 etc Can be done in a separate change too

alwin-joseph (approved):

user_guides/jakartaee/src/main/jbake/content/config.adoc
scottmarlow: Thanks, that does trickle down below as well for the numbering change.

scottmarlow: +1, I just renumbered the other below sections until we reach 5.5 :-), I'll include these changes along with the assertion updates.  I suspect that the assertion updates + doc changes can involve a few rounds of discussion + further changes so want to do that separately.

Issue #44, opened 2019-02-11
Servlet TCK secbasic#test7[_anno] depends on container specific behaviour


markt-asf:
The two tests in question are:

- `src/com/sun/ts/tests/servlet/spec/security/secbasic/Client.java#test7`
- `src/com/sun/ts/tests/servlet/spec/security/secbasic/Client.java#test7_anno`

The tests are identical apart from the second tests uses annotations rather than web.xml for configuration.
The tests use a servlet mapped to `/ServletSecTest`. Note this is an exact mapping, not a path mapping.
There is a security constraint also mapped to `/ServletSecTest`
The client requests `/ServletSecTest/j_security_check` without any credentials
The tests expects a 401.
The reasoning is explained at https://github.com/eclipse-ee4j/jakartaee-tck/blob/master/src/com/sun/ts/tests/common/jspservletsec/SecBasicClient.java#L340

I do not think that the specification supports this behaviour.

I think the specification requires the following:
Since exact mapping is used, the request does not match the servlet mapping.
This means the request is passed to default servlet since it has a mapping of `/*`.
The servletPath will be the empty string and the pathInfo will be `/ServletSecTest/j_security_check`.
This doesn't match any security constraints so the default servlet will try to server this resource.
The resource does not exist so a 404 is returned.

The test could be 'fixed' by making the servlet mapping `/ServletSecTest/*` but that does not address the issue of the underlying premise of the test being incorrect. Changing the expected output (and the explanatory comment) to expect 404 is one possible solution. Simply removing/disabling the tests is another.

I'm happy to try and produce a PR for this once there is consensus on how it should be handled.

janbartel:
I agree the test is flawed. At the very least the servlet and resource collection urls are wrong, and should have a trailing` '/*'`.  However, that said, I don't understand what is purported to be achieved with the request url containing a suffix of  '/j_security_check'.

markt-asf:
On reflection, I think looking for 404 is the correct way to handle this.
The test is checking that the addition of "/j_security_check" doesn't trigger something in the authentication mechanism.

arjantijms:
@markt-asf perhaps we could consider adding a test, for clarity, that a variant of 7 using "/blabla_check" instead of "/j_security_check" behaves exactly the same.

PR #558, opened 2020-10-27
Revert previous expected status code change which is invalidating HTTP BASIC authentication testing.

darranl:
The change to the status code has changed the test from performing a security test to performing a path availability test instead, this is making tests within "servlet/spec/security/secbasic/Client.java" invalid, specifically "test7" and "test7_anno".

"test7" starts with a comment describing the intent of this test, the intent is to test that the container denies access to the resource when accessed with an invalid username and password.  The text also describes that "j_security_check" is appended to the URL to verify this does not "trick" the server into allowing the request through.  At no point is "j_security_check" being described as being used as the trigger for authentication.

For "test7" the path "/servlet_sec_secbasic_web/ServletSecTest/j_security_check" is accessed, this matches against the constraint defined in the web.xml:

```
  <security-constraint>
    <web-resource-collection>
      <web-resource-name>MySecureBit0</web-resource-name>
      <url-pattern>/ServletSecTest</url-pattern>
      <http-method>GET</http-method>
      <http-method>POST</http-method>
    </web-resource-collection>
    <auth-constraint>
      <role-name>Administrator</role-name>
    </auth-constraint>
    <user-data-constraint>
      <transport-guarantee>NONE</transport-guarantee>
    </user-data-constraint>
  </security-constraint>
```

This constraint should be matched and enforced before the underlying container moves to indicating the presence (or not) of the specified resource.  Responding with a HTTP 404 is providing information that should not be available to an unauthenticated identity.

For "test7_anno" the servlet is annotated with:

```
@ServletSecurity(value = @HttpConstraint(EmptyRoleSemantic.DENY), httpMethodConstraints = {
    @HttpMethodConstraint(value = "GET", rolesAllowed = "Administrator"),
    @HttpMethodConstraint(value = "POST", rolesAllowed = "Administrator") })
@WebServlet(name = "ServletSecAnnoTestLogicalName", urlPatterns = {
    "/ServletSecAnnoTest" })
```

Again these constraints should be being matched BEFORE allowing the request through for processing.

In both cases the web.xml contains the following meaning HTTP BASIC authentication has been selected as the chosen authentication mechanism to challenge with when authentication is required:

```
  <login-config>
    <auth-method>BASIC</auth-method>
    <realm-name>default</realm-name>
  </login-config>
```

A HTTP status code of 401 is the correct status code to challenge the remote client with after the failed authentication.

Revert "Correct the response code being looked for."

This reverts commit 947f34a3bd9055c599d9ccfa5239eff68afe9957.

Signed-off-by: Darran Lofthouse <darran.lofthouse@jboss.com>

### DIFF
--- a/src/com/sun/ts/tests/common/jspservletsec/SecBasicClient.java
+++ b/src/com/sun/ts/tests/common/jspservletsec/SecBasicClient.java
@@ -365,8 +365,9 @@ public class SecBasicClient extends BaseUrlClient {
     TEST_PROPS.setProperty(BASIC_AUTH_USER, "invalid");
     TEST_PROPS.setProperty(BASIC_AUTH_PASSWD, password);
     // The servlet is mapped to "/ServletSecTest" so this request should result
-    // in a 404 since no Servlet is mapped to the requested URI.
-    TEST_PROPS.setProperty(STATUS_CODE, NOT_FOUND);
+    // in a 404 since no Servlet is mapped to the requested URI or 401 if the container
+    // rejects the incoming BASIC header.
+    TEST_PROPS.setProperty(STATUS_CODE, UNAUTHORIZED + "," + NOT_FOUND);
     invoke();
 
     dumpResponse(); // debug aid
--- a/src/com/sun/ts/tests/common/webclient/validation/WebValidatorBase.java
+++ b/src/com/sun/ts/tests/common/webclient/validation/WebValidatorBase.java
@@ -171,28 +171,43 @@ public abstract class WebValidatorBase implements ValidationStrategy {
       return true;
     }
 
-    // test status code not null, compare with that of the response
-    if (sCode.charAt(0) != '!') {
-      if (!sCode.equals(resCode)) {
-        TestUtil.logErr("[WebValidatorBase] Unexpected Status Code "
-            + "recieved from server.  Expected '" + sCode + "' received '"
-            + resCode + "'");
-        return false;
-      }
+    /*
+     * Take sCode as a comma separated list of status codes.
+     *
+     * If prefixed by "!" the response status code must not match any in the list.
+     *
+     * Otherwise matching any in the list is accepted.
+     */
 
-      TestUtil.logTrace("[WebValidatorBase] Expected Status Code '" + sCode
-          + "' found in response line!");
+    boolean exclusions = sCode.charAt(0) == '!';
+    String[] sCodes = exclusions ? sCode.substring(1).split(",") : sCode.split(",");
+
+    if (exclusions) {
+        for (String current : sCodes) {
+            if (current.equals(resCode)) {
+                TestUtil.logErr("[WebValidatorBase] Unexpected Status Code "
+                    + "recieved from server.  Expected any value except '" + sCode
+                    + "', received '" + resCode + "'");
+                return false;
+            }
+        }
     } else {
-      sCode = sCode.substring(1);
-      if (sCode.equals(resCode)) {
-        TestUtil.logErr("[WebValidatorBase] Unexpected Status Code "
-            + "recieved from server.  Expected any value except '" + sCode
-            + "', received '" + resCode + "'");
-        return false;
-      }
+        boolean found = false;
+        for (String current : sCodes) {
+            if (current.equals(resCode)) {
+                TestUtil.logTrace("[WebValidatorBase] Expected Status Code '" + current
+                        + "' found in response line!");
+                found = true;
+                break;
+            }
+        }
 
-      TestUtil.logTrace("[WebValidatorBase] Status Code '" + sCode
-          + "' not found in response line!");
+        if (!found) {
+            TestUtil.logTrace("[WebValidatorBase] Status Code '" + sCode
+                      + "' not found in response line!");
+
+            return false;
+        }
     }
 
     return true;


arjantijms:
I'm not entirely sure this is the case. The pattern that's protected is:

```xml
<url-pattern>/ServletSecTest</url-pattern>
```

Which is an **exact mapping**. This means `ServletSecTest` and only `ServletSecTest` has this constraint. `ServletSecTest/j_security_check` is another path, that does not have this constraint.

In Piranha I passed the revised test without having to do anything special. For GlassFish I remembered that `j_security_check` triggered the authentication check because of a bug. `AuthenticatorBase#processSecurityCheck` is shared by the BASIC and FORM mechanisms, and looks at the `j_security_check` path:

```java
 private int processSecurityCheck(HttpRequest hrequest, HttpResponse hresponse,  LoginConfig config) throws IOException {

        // Special handling for form-based logins to deal with the case
        // where the login form (and therefore the "j_security_check" URI
        // to which it submits) might be outside the secured area
        String contextPath = this.context.getPath();
        String requestURI = hrequest.getDecodedRequestURI();
        if (requestURI.startsWith(contextPath) && requestURI.endsWith(FORM_ACTION)) {
            if (!authenticate(hrequest, hresponse, config)) {
                if (log.isLoggable(FINE)) {
                    String msg = " Failed authenticate() test ??" + requestURI;
                    log.log(FINE, neutralizeForLog(msg));
                }
                return END_PIPELINE;
            }
        }
	return INVOKE_NEXT;
    }
```

There it starts an `authenticate()` action, which for the semantics of an unconstrained path with a BASIC authentication mechanism, should not be performed.

darranl:
Thank you @arjantijms I see what you mean.

I think this is still leaving the question is "Should /j_security_check be taken into account during the pattern matching for the auth-constraint?"

The following is the comment from the top of "test7":

> If user has not been authenticated and user attempts to access a protected web resource+<j_security_check> , and user enters an invalid username and password, the container denies access to the web resource.

With the change to 404 the container is not demonstrating that it has denied access to the web resource, instead it is demonstrating that it has not matched against an auth-constraint and instead reported the resource does not exist.  But the text in the test appears to be assuming it would be matching against the web resource.

So for example removing "/j_security_check" leaves us with "/ServletSecTest" which matches a defined constraint, which on unsuccessful authentication leads to denied access.

It is feeling like the original test was written based on an assumption about "/j_security_check" being omitted from the pattern matching as a special path even though this is not mentioned in the specification.  

But that would leave the question, what it "test7" actually testing?  It is no longer testing access to a resource is denied.  So instead it is testing two other behaviours:

- The trailing j_security_check is ignored.
- An invalid BASIC authentication header is ignored.

Both of these behaviours are also not defined in the specification.

arjantijms:
> I think this is still leaving the question is "Should /j_security_check be taken into account during the pattern matching for the auth-constraint?"

Nope, I don't think it can be taken into account, since it's not part of the security-constraint. There's nothing special about `/j_security_check` wrt the pattern matching rules (either for Servlet dispatching, or for the `security-constraint` parsing). It's just a string really that the FORM authentication mechanism (and only the FORM one) happens to look for.

> It is feeling like the original test was written based on an assumption about "/j_security_check" being omitted from the pattern matching as a special path even though this is not mentioned in the specification.

I don't know, I found quite a lot of tests being of somewhat low quality. Tests are copy/pasted around a lot, and many tests in the Servlet TCK contain things that are completely irrelevant for what they are testing, even up to extra Servlets with methods that are never used.

I suspect in this case the test is just based on the GF bug, where the base class code happens to pick up `j_security_check `, even though this is only intended for FORM.

stuartwdouglas:
I agree with @arjantijms , I think the current behaviour is correct as per spec. 

The issue is though that pretty much all containers that pass the TCK will have taken the lead from the TCK and implemented it https://github.com/undertow-io/undertow/blob/ed4d901174cc93847bfabd903892dc02dfc28e19/servlet/src/main/java/io/undertow/servlet/handlers/security/ServletAuthenticationConstraintHandler.java#L46 (sometimes it is easier to just do what the TCK is testing rather than challenging the spec).

If other containers have done the same thing then I think we should probably revert the commit and just change the spec.

stuartwdouglas:
CC @gregw

darranl:
Ok, if everyone is happy the behaviour is as expected I will submit a PR to adjust the test comment, the test is now no longer testing access to the resource is denied - instead it is now testing the request does not map to the resource and does not trigger the security constraint of the resource.

It does feel odd that j_security_check testing was getting involved in HTTP BASIC authentication testing but I agree this does feel like it was testing a specific bug somewhere rather than testing compliance with the specification.

I am running into a separate issue that the TCK is expecting a web application configured to support HTTP BASIC authentication ignores the incoming HTTP header but that is a separate issue to the path matching.

darranl:
Still thinking about the comment from Stuart.

> The issue is though that pretty much all containers that pass the TCK will have taken the lead from the TCK and implemented it

The main problem with this change is it is not insignificant in a release that is supposed to be about the package change and removal of optional specifications.  Once a new requirement is placed on container to ignore the j_security_check suffix it introduces the new problem of defining the times the container does need to handle the j_security_check.

Verifying a container does not allow unauthenticated / unauthorized access to a secured resource feels a valid scenario in a TCK test.  Forcing a containers security layer to not activate despite three potential triggers for it to activate seems beyond the scope of what should be tested.

markt-asf:
Tomcat passes the revised test. It was Tomcat failing the original test that identified the faulty test and was the trigger for the TCK to be fixed.

Thanks to @arjantijms for tracking down the root cause of the faulty test.

I am strongly against introducing any sort of special handling for `j_security_check` in the path matching rules.

Given that the TCK has been used in the past to determine the correct behaviour when the spec was ambiguous, I have some sympathy for implementations that previously adjusted to pass the faulty version of this test.  I would have no objection to this test (in its current form) being added to the exclude list for Jakarta EE 9 with the expectation it is removed from the exclude list for the next release.

darranl:
Just to be clear on the WildFly side, the main issue is not so much the handling of "j_security_check" rather the fact that the test now mandates that we ignore the BASIC authentication header in a web application configured to use BASIC authentication.

gregw:
I'm a bit torn by this one.  I agree with @arjantijms and @stuartwdouglas that we should not be treating j_security_check generally, specially when we are using other authentication mechanisms.
However, I also see @darranl point about changing behaviour in a release that is just meant to be a rename.

So how about we change the test so that either a 401 or a 404 is an acceptable response?  The main thing is that we shouldn't get a resource. 

We can then tighten up the specification and TCK in a later release.

markt-asf:
@darranl ah, understood. Since the request doesn't match a security constraint this is getting into the area of preemptive authentication. I did some work on this for Tomcat a while back and recall digging into various specs to try and figure out what should happen in the case of invalid credentials provided for a resource that did not need any. I don't recall if there was a definitive answer. Let me try and find mt notes from that work...

janbartel:
Just to remind everyone this is an open issue , but does not appear to have been referenced on any of the PRs to date: https://github.com/eclipse-ee4j/jakartaee-tck/issues/44

IMHO the correct response is a 404 and the test should reflect that, but as @gregw and others say, many other containers will have coded to pass the previous version of the test which expects a 401. So, just for this release (which is rename only, no behaviour changes), then we should accept 404 and 401 (with javadoc explanations as to why), but the next release _should_ fix the test to expect a 404, and add another test, if necessary, to be in accordance with the purported aim of the test. The next spec may need to clarify this as well.

darranl:
Thanks for the link to the issue, that is very useful.

As the discussion continues I will update the PR to check 401 and 404 so if that is accepted for now it is ready.

markt-asf:
I went back over my notes and the short answer is that the expected behaviour for preemptive authentication is (in my view) undefined. Others held the view that if credentials are provided they must be processed.

[RFC 7235](https://tools.ietf.org/html/rfc7235#section-3.1) defines a 401 response as: 

> the request has not been applied because it lacks valid authentication credentials for the target resource

The credentials are invalid but the resource doesn't require any credentials. I think you could use the quote above to argue for and against a 401 in this case.

RFC 7235 goes on to state:

> The server generating a 401 response MUST send a WWW-Authenticate header field (Section 4.1) containing at least one challenge applicable to the target resource.

Here you could argue that no challenges are applicable to the resource (hence the requirement is impossible to meet which implies the credentials should have been ignored in the first place) or you could argue that given the user is trying to provide credentials, providing the appropriate challenge for the current realm is reasonable.

Another quote from RFC7235:

> Upon receipt of a request for a protected resource that omits credentials, contains invalid credentials (e.g., a bad password) or partial credentials (e.g., when the authentication scheme requires more than one round trip), an origin server SHOULD send a 401 (Unauthorized) response that contains a WWW-Authenticate header field with at least one (possibly new) challenge applicable to the requested resource.

You need to parse the above carefully but if you do it only applies to protected resources. Behaviour for unprotected resources is undefined.

In Tomcat we made preemptive authentication optional. When disabled (the default) credentials are only examined when resources are protected. When enabled, provided credentials are processed as if the resource is protected (and a 401 sent if they are invalid).

How we handle preemptive authentication probably needs a new issue raised against the Servlet spec.

gregw:
@markt-asf Thanks for all those quotes and I broadly agree that they can be argued both ways.  

But I do not see how any request to `j_security_check` can be views as a preemptive authentication in a webapp that has not been configured to accept FORM authentication.
So we need to step back even further and decide if a webapp that does not support FORM auth for a request to `/foobar/j_security_check`:
 - Does that represent a request with credentials (invalid or otherwise)?
 - Is `/foobar` the target resource? 

Note also that I think that FORM authentication (and many other auth mechanisms) fall outside of RFC7235 2.1, as that is really just for BASIC, DIGEST etc. and not for auth mechanisms like FORM that use redirection and/or other mechanisms.

I think 404 is the correct response, but am OK with a 401 as well as a grandfathered response in the TCK

darranl:
@gregw I think in this case the use of j_security_check is over complicating the test as everyone keeps referring back to this specific point but there is a second trigger for authentication in the test.

The test uses a web application configured for HTTP BASIC authentication and the test submits a BASIC authentication header which contains an invalid password.  It may not be the only option as described by @markt-asf but IMO it is a valid option that the container turns the request around with a HTTP 401 after identifying the incoming credentials were invalid.

As has been discussed if the test accepts a 401 and 404 that would allow the container to handle the authentication header.

But as has been pointed out in a couple of comments this test is still incomplete so would be good to revisit after Jakarta EE 9 as well as reviewing if some of these scenarios should be refined in the specification.

scottmarlow:
> As has been discussed if the test accepts a 401 and 404 that would allow the container to handle the authentication header.

@markt-asf do you agree with @darranl updating the pull request to accept either a 401 or 404?  I didn't expect that we would be updating this test now (we plan to promote the Platform TCK possibly on Friday).  We could still merge a change to this test if you or @stuartwdouglas approve (as leads of the Servlet SPEC).  

As I see this, @arjantijms has pointed out the expected behaviour from a security point of view, for which checking for both HTTP 404 or (grandfathered) 401 will handle.  With a possible future Servlet SPEC change for how we handle preemptive authentication (if someone wants to raise a servlet issue for that).

gregw:
@darranl 

I think that:
 - a request to `pageSec` with invalid BASIC credentials should return a 401 and only a 401
 - a request to `pageSec + "j_security_check" with or without BASIC credentials should return a 404, but a 401 can be acceptable because of history.

markt-asf:
@scottmarlow Yes I'd be happy with a PR that updated the test to accept either a 401 or 404. I'd also be happy with excluding the test.

darranl:
> a request to `pageSec + "j_security_check" with or without BASIC credentials should return a 404, but a 401 can be acceptable because of history.

That would effectively be saying the test is banning proactive authentication of the incoming security header.

However once we start to combine the use of the HttpServletRequest.authenticarte(HttpServletResponse) method we end up in a scenario where we need to be processing incoming security headers again when targeting a resource not backed by a security constraint.

darranl:
I have submitted a replacement PR which adjusts the test to accept both 401 and 404, going to try and build the TCK locally to also run the test here.

scottmarlow:
> I have submitted a replacement PR which adjusts the test to accept both 401 and 404, going to try and build the TCK locally to also run the test here.

Hi Darran, instead of building the entire Platform TCK, you could instead rebuild just the servlet TCK tests (or a subset by changing into the relevant test folders).  It should just be a few minutes to test instead of hours.  I'll look for an example.

scottmarlow:
@darranl https://gist.github.com/scottmarlow/f7b4a166f21c3d89ed471d56d5c41003 is a bit more than you want, so will copy the needed parts here:

```
cd src/com/sun/ts/tests/servlet
OR 
cd src/com/sun/ts/tests/servlet/spec/security
ant compile build
# note that files under the TCK dist folder that contains test archives will be updated
```

You need to do the `ant compile build` under the folder that contains the test classes.  I hope that makes sense.

gregw:
> > a request to `pageSec + "j_security_check"` with or without BASIC credentials should return a 404, but a 401 can be acceptable because of history.
> 
> That would effectively be saying the test is banning proactive authentication of the incoming security header.

I don't think so.  The webapp uses BASIC auth, so if a client wants to proactively authenticate then they just send a request to `pageSec` with their credentials.
The URI `pageSec + "j_security_check"`  has nothing to do with the secured resource when BASIC is the auth method.   I think it is unreasonable to require a container to check for FORM auth requests if it has not been configured with FORM auth.

darranl:
@gregw Your comment is still focusing on the FORM auth aspects, that is not relevant to what I am saying. j_security_check is not a part of the trigger I am talking about.  

The trigger I am talking about it sending a HTTP BASIC header to a web application which is also secured to use BASIC authentication.  

As the web application is configured to support HTTP BASIC authentication the header is processed by the authentication mechanisms as the request is inbound and as the header contains an invalid password it is turned round and returned back to the client.

This would be just as applicable for a request to /pathspec/somesubpath.

gregw:
> This would be just as applicable for a request to /pathspec/somesubpath.

Ah got you.   

So we need to be careful with this kind of test. We don't want a client to be able to probe `/some/private/path/some/file` and use the 401 vs 404 returned to determine what files do or do not exist within `/some/private/path`.

However, in the case of this test, the secure resource is `/ServletSecTest`, so I think that a request to `/ServletSecTest/anything` is not secured.  If the resource exists it will be returned, if it does not exist then a 404 should be returned.  A 401 should never be returned...... except in the grandfathered case of `/ServletSecTest/j_security_check` because some containers see that as a request of `/ServletSecTest`

So for this test, it is still correct that either 404 or 401 is acceptable.... but it is ONLY because the path testing is `/ServletSecTest/j_security_check` and `/ServletSecTest/anythingelse` should return a 404 if the resource does not exist.  If the protected pattern was  `/ServletSecTest/*` then 401 should always be returned regardless if the resource exists or not.

darranl:
This is the piece that I think needs some ongoing discussion

> /ServletSecTest/anythingelse should return a 404 if the resource does not exist.

I think this is worth discussing hopefully in the next round of spec discussions regarding how pro-active auth fits with the spec.  IMO this ties in very much with support for HttpServletRequest.authenticarte(HttpServletResponse) which in itself can trigger security headers coming in against resources that do not have a defined security constraint.

scottmarlow:
I started a test run via https://ci.eclipse.org/jakartaee-tck/job/jakartaee-tck-scottmarlow/job/darranl-unauthorized_status_code/2/

I also am using a custom GlassFish 6.0 build https://ci.eclipse.org/glassfish/job/glassfish_build-and-test-using-jenkinsfile/job/PR-23264/1/artifact/bundles/glassfish.zip as I wanted to see if improves TCK testing reliability.

stuartwdouglas:
IMHO either a 404 or 401 should be acceptable here. The spec does not say how this should be handled (i.e. the spec does not say anything about proactive auth one way or another), so both are valid depending on how the container is configured.

scottmarlow (approved):

stuartwdouglas (approved):

arjantijms:
@markt-asf 
> I went back over my notes and the short answer is that the expected behaviour for preemptive authentication is (in my view) undefined.

Indeed, preemptive authentication in just the Servlet spec is undefined.

If I got my history correct, the Servlet spec at some point had a more concrete specification of the authentication mechanism concept on its agenda, which tied in with custom authentication mechanisms.

At some point the powers that be saw universal applicability of this concept and extracted it to a separate spec intended to provide these authentication mechanisms for various protocols/APIs, including JMS, Remote EJB and SOAP. The Servlet parts of that were captured in the Servlet Container Profile, and with much delay this "JMAC" spec was published as JASPIC.

Instead of clarifying the authentication mechanism in the Servlet spec, the Servlet spec was just updated with text that it was recommended that all Servlet containers implement this Servlet Container Profile.

The problems here are twofold; "recommended" is quite unfortunate wording for something that was originally going to be directly in the Servlet spec, and, it only speaks about "additional authentication mechanisms". It says nothing about whether or not the semantics of these apply to the build-in authentication mechanisms as well.

If you ask me a historic mistake and the start of my 10+ years journey to convince Greg (but that latter is besides the point here 😝)

gregw:
On Thu, 29 Oct 2020, 14:08 Arjan Tijms, <notifications@github.com> wrote:

> If you ask me a historic mistake and the start of my 10+ years journey to
> convince Greg (but that latter is besides the point here 😝).
>

Well it's not really besides the point here... because as well as fixing
little issues, unless we start thinking about bigger improvements now, then
we'll never have time to include them in any version of the spec.

With regards to authentication and authorisation we have a very minimal
solution supported on servlets that is well short of common/best practise.
Then in JASPIC we have a fully (or at least more) functional solution that
as far as I can tell nobody ever uses. We've supported it in jetty for 10+
years but I've never seen a single use/bug/question about it.

I'm very receptive about making a fully functional authentication &
authorisation framework a part of the servlets standard it at least define
the integration as part of the standard. It would be hard to do worse than
the status quo.... But I'm sure we're up to the challenge :)

Cheers

>

darranl:
> I'm very receptive about making a fully functional authentication &
> authorisation framework a part of the servlets standard 

+1 would definitely be involved in discussions around that - things have moved on a long way since the first four authentication mechanisms were selected for the Servlet specification.

arjantijms:
> With regards to authentication and authorisation we have a very minimal solution supported on servlets that is well short of common/best practise. Then in JASPIC we have a fully (or at least more) functional solution that
as far as I can tell nobody ever uses.

Oh, but it is used. Jakarta EE Security directly builds on it, and on it various implementations of for example MP JWT build  on that (including SmallRye from Red Hat). 

But it's not just JASPIC (now Jakarta Authentication) that's being asked for directly by users, but us here asking for exactly the semantics that its Servlet Container Profile specified. It's a bit like going in circles:

We have an unspecified issue in Servlet and users or us ask about it. Exactly that issue is specified in the Servlet Container Profile. But we don't adhere to the Servlet Container Profile, since nobody ever asked for it?

Essentially adopting the semantics of the Authentication Servlet Container Profile, and stating the existing four authentication mechanisms adhere to it, would solve 80% of the issues, if not more. Higher level security frameworks (Soteria, Shiro, Elytron, you name them) can plug in to that for the more advanced use cases. This is also exactly what I'm doing for the Piranha Servlet implementation.

I think it was Stuart (not 100% sure) who suggested to do a few simplifications to the Authentication Servlet Container Profile.

arjantijms:
>  IMO this ties in very much with support for HttpServletRequest.authenticarte(HttpServletResponse) which in itself can trigger security headers coming in against resources that do not have a defined security constraint.

This is btw one of those small things where the Authentication Servlet Container Profile exactly solves the problem. It simply states that after a call to `HttpServletRequest.authenticate` authentication is mandatory.

darranl:
> I think it was Stuart (not 100% sure) who suggested to do a few simplifications to the Authentication Servlet Container Profile.

Maybe Stuart said something as well but we did have a discussion a while back, it feels like the key piece of the JASPIC specific that is needed is the servlet profile so could be an opportunity to have a slimmed option for the servlet profile.

> This is btw one of those small things where the Authentication Servlet Container Profile exactly solves the problem. It simply states that after a call to HttpServletRequest.authenticate authentication is mandatory.

I don't think it is quite covering it, the issue is HttpServletRequest.authenticate is performed on one incoming request then the authentication headers as a result of the challenge come in on a second independent request.  The resource that called the method may not be under a security constraint but now has incoming security headers

arjantijms:
> Maybe Stuart said something as well but we did have a discussion a while back, it feels like the key piece of the JASPIC specific that is needed is the servlet profile so could be an opportunity to have a slimmed option for the servlet profile.

Yes absolutely. Note that the Servlet spec already mentions that it's only about the servlet profile, not full Jakarta Authentication:

> To facilitate portable implementation and integration of additional container authentication mechanisms, it is recommended that all servlet containers implement the Servlet Container Profile of The Jakarta Authentication Specification.

https://jakarta.ee/specifications/servlet/5.0/jakarta-servlet-spec-5.0#additional-container-authentication-mechanisms

Frankly, since Jakarta Authentication was historically extracted from the work in progress Servlet spec and had extra stuff added, I wonder if we should not just fold the Servlet parts after simplification back into Servlet, and declare Jakarta Authentication itself a stable spec. The SOAP and JAAS bridge things are certainly not going to be updated again, and the JMS, EJB etc things never came off the ground. 

Just thinking out loud here though, maybe there's a better solution.

darranl:
> Yes absolutely. Note that the Servlet spec already mentions that it's only about the servlet profile, not full Jakarta Authentication:

I think one of the issues I ran into was the servlet profile is only interested in the server side of the integration but the general JASPIC APIs required both the client and server sides to be implemented.

> Frankly, since Jakarta Authentication was historically extracted from the work in progress Servlet spec and had extra stuff added, I wonder if we should not just fold the Servlet parts after simplification back into Servlet, and declare Jakarta Authentication itself a stable spec.

Does sound reasonable.

arjantijms:
> I don't think it is quite covering it, the issue is HttpServletRequest.authenticate is performed on one incoming request then the authentication headers as a result of the challenge come in on a second independent request. The resource that called the method may not be under a security constraint but now has incoming security headers

I hear you. There's various things to consider here though

1. `HttpServletRequest.authenticate` mandates security in the current request as per the Authentication Servlet profile.
2. This means the configured authentication mechanism is called, and it **must** try to authenticate. Indeed, this means sending a challenge or processing headers (or any form of credentials) in the current request.
3. Depending on the authentication mechanism, an authentication dialog may now be in progress, where the authentication mechanisms when called in the next request is aware of it previously sending a challenge. Clearly BASIC is not specified as such an authentication mechanism.
4. The resource that initially called `HttpServletRequest.authenticate` and caused the challenge to be sent, can call `HttpServletRequest.authenticate` again on the second independent request. This will force, as per item 2, those incoming security headers to be processed.
5. In parallel to this, the Authentication Servlet profile does specify an authentication mechanism is **always** given the opportunity to authenticate before the resource is invoked, whether that resources is constrained or not (preemptive authentication). It depends on the authentication mechanism itself in that case, not on the overarching security machinery, if it wants to do that.

So essentially we have it 90% covered. The one thing that is indeed not covered, if I'm not mistaken, is what you meant:

6. The container remembers that on request 1 `HttpServletRequest.authenticate` was called, making authentication mandatory and causing a challenge to be sent, and that now on independent request 2 the authentication mechanism is called **before** the resource is invoked with a mandatory flag.

arjantijms:
> I think one of the issues I ran into was the servlet profile is only interested in the server side of the integration but the general JASPIC APIs required both the client and server sides to be implemented.

Yes, absolutely. That's one of those things that may have seen like a good idea at the time and given that e.g. the MP Rest Client and JAX-RS client builder came to fruition later had indeed some theoretical merit, but in the Servlet Profile it doesn't make any sense. IMHO, in a Jakarta Authentication Lite, or when folded back into Servlet (or whatever we choose), we should absolutely omit these.

arjantijms:
p.s.

@scottmarlow regarding the update of this test itself, isn't it an issue that Servlet 5.0 has already passed the ballot and has been fully released?

darranl:
I think it is covered mostly by points 4 and 5.  I wouldn't be in favour of 6 as that has it's own issues.

The preemptive authentication under step 5 I think works well to intercept the headers if they are inbound but that then directly feeds into the discussion which has been at the core of this PR which is can the container turn the request back round with a 401 if the validation fails or if a further round trip is needed.

kwsutter:
> p.s.
> 
> @scottmarlow regarding the update of this test itself, isn't it an issue that Servlet 5.0 has already passed the ballot and has been fully released?

Yes.

scottmarlow:
> > p.s.
> > @scottmarlow regarding the update of this test itself, isn't it an issue that Servlet 5.0 has already passed the ballot and has been fully released?
> 
> I'm going to modify my answer a bit... The Servlet 5.0 ballot was for a specific version of the Compatible Impl and the 5.0.0 TCK. That's still valid.
> 
> If this change gets accepted into the Platform TCK, then that would necessitate an updated 5.0.1 TCK (I'd go with a patch release since you're just expanding the acceptable answers). That would require re-testing by the 5.0 CI. This wouldn't have to be done before Jakarta EE 9 completes, but it should be shortly there-after. We'd have to run this idea by the Spec Committee per the [TCK process](https://jakarta.ee/committees/specification/tckprocess/): _"Projects will submit the EFTL proposed final binary of the TCK for approval to the Specification Committee."_
> 
> Thus, the safest route at this point would be to exclude the test for now. And, then update both the Platform TCK and Servlet TCK post Jakarta EE 9.

@kwsutter with the safest route, I am fine with excluding the test for now, however, we would have to wait until EE 9.1 to enable the test again which I think you meant but want to confirm.

Also, just to confirm, the Jakarta EE 9 Platform TCK will exclude the test but the Standalone Servlet TCK (5.0.0) will continue to include the test for now.  Until EE 9.1, for which we will merge this pull request and enable the test again.  Sound right?

We will add exclude entries for:
- test=com/sun/ts/tests/servlet/spec/security/secbasic/Client.java\#test7
- test=com/sun/ts/tests/servlet/spec/security/secbasic/Client.java\#test7_anno

stuartwdouglas:
This has gone off topic a little bit.

In terms of pro-active auth we should definitely add something to the spec, but IMHO it should just mention that containers have the option of doing auth proactivly or not, based on container config. We could add a setting to web.xml I guess, but I don't have strong feelings either way on that one. WildFly does support this through jboss-web.xml and users do use it, so maybe it is worth supporting. 

In terms of authentication mechanisms from a Quarkus POV there are two main reasons why JASPIC is not really suited to our needs:

- It is based on Servlet and Servlet is not a required part of our HTTP stack. e.g. if a user is just using JAX-RS then we just run JAX-RS on top of Vert.x without a Servlet layer in between. If we had a minimal standard that just defined a HTTP request and response then we could do common auth based on this, but we are not going to require Servlet. This is also an issue for WildFly as well to some extent, our admin console also needs authentication but does not want to depend on Servlet.

- The authentication mechanisms are inherently blocking, so they can't be re-used for non blocking services that want to do their work on the IO thread. Maybe loom will make this less of an issue, but  I am not sure.

arjantijms:
> It is based on Servlet and Servlet is not a required part of our HTTP stack.

I hear what you're saying, but it's not entirely the case. Jakarta Authentication itself is completely API neutral. Nothing in it depends or is based on Servlet (or Enterprise beans (EJB) or Jakarta Messages (JMS) for that matter.

Just look at the pom: it has no dependencies at all: https://github.com/eclipse-ee4j/authentication/blob/master/api/pom.xml

Now what it does have is profiles in the spec defining semantics for different APIs. It has profiles for Servlet, SOAP and JAAS, with planned ones for JMS and EJB, but those failed to get off the ground as mentioned.

> If we had a minimal standard that just defined a HTTP request and response then we could do common auth based on this, but we are not going to require Servlet.

Yes, that was what we talked about on the Servlet list. Concerning Quarkus (and other implementations without Servlet) there's two options that could be considered:

* Go for that minimal standard with just approximately the HTTP request and response, which may be enough for the Authentication Servlet Profile (existing one, or "lite" one if needed)
* "Simply" define a profile covering Jakarta REST (JAX-RS). A profile here would be a few pages of text, my guess would be a copy of the Servlet text, with some adjustments where needed for Jakarta REST.

darranl:
Regarding:

> Go for that minimal standard with just approximately the HTTP request and response

This was also the problem we ran into with WildFly Elytron, we don't really want our own SPIs for authentication mechanisms but needed to run both with and without the servlet container.  A minimal specification would also help greatly in that regard.

Going further off topic is the client side of this problem.

scottmarlow:
> Also, just to confirm, the Jakarta EE 9 Platform TCK will exclude the test but the Standalone Servlet TCK (5.0.0) will continue to include the test for now. Until EE 9.1, for which we will merge this pull request and enable the test again. Sound right?
> 
> We will add exclude entries for:
> 
>     * test=com/sun/ts/tests/servlet/spec/security/secbasic/Client.java#test7
> 
>     * test=com/sun/ts/tests/servlet/spec/security/secbasic/Client.java#test7_anno

^ is excluded via https://github.com/eclipse-ee4j/jakartaee-tck/pull/562 with a TODO added as a reminder to remove the exclude after merging this pr in the next EE release (e.g. Jakarta EE 9.1 or whatever the number is).  I will label this pull request as `future`.

stuartwdouglas (approved):

scottmarlow:
We have not yet received a [Servlet TCK challenge](https://github.com/eclipse-ee4j/servlet-api/issues/new) for generating a 5.0.1 TCK.  If we do and the challenge is accepted, we will build a 5.0.1 Standalone Servlet TCK.  This addresses the comments above about a 5.0.1 TCK.  

We will soon merge this pull request for 9.1 and remove the test excludes:
```
install/jakartaee/bin/ts.jtx
+# Excluding this test until we can merge https://github.com/eclipse-ee4j/jakartaee-tck/pull/558
+# TODO: In the next release after Jakarta EE 9 (perhaps Jakarta EE 9.1 or 10) the pull request will be merged and this exclude removed.
+com/sun/ts/tests/servlet/spec/security/secbasic/Client.java#test7
+com/sun/ts/tests/servlet/spec/security/secbasic/Client.java#test7_anno

install/servlet/bin/ts.jtx
+# Excluding this test until we can merge https://github.com/eclipse-ee4j/jakartaee-tck/pull/558
+# TODO: In the next release after Jakarta EE 9 (perhaps Jakarta EE 9.1 or 10) the pull request will be merged and this exclude removed.
+com/sun/ts/tests/servlet/spec/security/secbasic/Client.java#test7
+com/sun/ts/tests/servlet/spec/security/secbasic/Client.java#test7_anno
```

scottmarlow:
We are getting closer to finishing 9.1 but there is still time to remove the test excludes and this change, so updated the label to `9.1`.

scottmarlow:
Sorry @darranl that didn't have time to coordinate this change for EE 9.1

PR #567, opened 2020-11-01
Fix SessionCookieConfig constructor test

Thihup:
In the TestListener of SessionCookieConfig [1], if there is an unexpected value, it adds to the comment a value of "FAILED-something". However, the test servlet searches for "-FAILED-", so none of the validations actually are checked.

As I can tell most of the containers (Jetty, Liberty, Undertow) fail the getName() test, usually returning _JSESSIONID_ if it wasn't previously set where it should be null.

[1] https://github.com/eclipse-ee4j/jakartaee-tck/blob/master/src/com/sun/ts/tests/servlet/api/jakarta_servlet_http/sessioncookieconfig/TestListener.java

### DIFF
--- a/src/com/sun/ts/tests/servlet/api/jakarta_servlet_http/sessioncookieconfig/TestListener.java
+++ b/src/com/sun/ts/tests/servlet/api/jakarta_servlet_http/sessioncookieconfig/TestListener.java
@@ -47,48 +47,48 @@ public class TestListener implements ServletContextListener {
     scf.setSecure(isSecure);
 
     if (!scf.getComment().equals(comment.toString())) {
-      comment.append("|getCommentFAILED-expecting-" + comment + "-got-"
+      comment.append("|getComment-FAILED-expecting-" + comment + "-got-"
           + scf.getComment());
       scf.setComment(comment.toString());
     }
 
     if (!scf.getPath().equals(path)) {
       comment
-          .append("|getPathFAILED-expecting-" + path + "-got-" + scf.getPath());
+          .append("|getPath-FAILED-expecting-" + path + "-got-" + scf.getPath());
       scf.setComment(comment.toString());
     }
 
     if (!scf.isSecure()) {
       comment.append(
-          "|isSecureFAILED-expecting-" + isSecure + "-got-" + scf.isSecure());
+          "|isSecure-FAILED-expecting-" + isSecure + "-got-" + scf.isSecure());
       scf.setComment(comment.toString());
     }
     if (scf.isHttpOnly()) {
-      comment.append("|isHttpOnlyFAILED-expecting-" + httpOnly + "-got-"
+      comment.append("|isHttpOnly-FAILED-expecting-" + httpOnly + "-got-"
           + scf.isHttpOnly());
       scf.setComment(comment.toString());
     }
     if (!scf.getDomain().equals(domain.toString())) {
       comment.append(
-          "|getDomainFAILED-expecting-" + domain + "-got-" + scf.getDomain());
+          "|getDomain-FAILED-expecting-" + domain + "-got-" + scf.getDomain());
       scf.setComment(comment.toString());
     }
 
     if (scf.getMaxAge() != maxage) {
       comment.append(
-          "|getMaxAgeFAILED-expecting-" + maxage + "-got-" + scf.getMaxAge());
+          "|getMaxAge-FAILED-expecting-" + maxage + "-got-" + scf.getMaxAge());
       scf.setComment(comment.toString());
     }
 
     if (scf.getName() != null) {
-      comment.append("|getNameFAILED-expecting-null-got-" + scf.getName());
+      comment.append("|getName-FAILED-expecting-null-got-" + scf.getName());
       scf.setComment(comment.toString());
     }
 
     scf.setName(name);
     if (!scf.getName().equals(name)) {
       comment
-          .append("|getNameFAILED-expecting-" + name + "-got-" + scf.getName());
+          .append("|getName-FAILED-expecting-" + name + "-got-" + scf.getName());
       scf.setComment(comment.toString());
     }
   }
--- a/src/com/sun/ts/tests/servlet/api/jakarta_servlet_http/sessioncookieconfig/TestListener.java
+++ b/src/com/sun/ts/tests/servlet/api/jakarta_servlet_http/sessioncookieconfig/TestListener.java
@@ -42,7 +42,6 @@ public class TestListener implements ServletContextListener {
     scf.setComment(comment.toString());
     scf.setDomain(domain);
     scf.setHttpOnly(httpOnly);
-    scf.setMaxAge(maxage);
     scf.setPath(path);
     scf.setSecure(isSecure);
 
@@ -74,6 +73,12 @@ public class TestListener implements ServletContextListener {
       scf.setComment(comment.toString());
     }
 
+    if (scf.getMaxAge() != -1) {
+      comment.append("|getMaxAge-FAILED-expecting-(-1)-got-" + scf.getMaxAge());
+      scf.setComment(comment.toString());
+    }
+
+    scf.setMaxAge(maxage);
     if (scf.getMaxAge() != maxage) {
       comment.append(
           "|getMaxAge-FAILED-expecting-" + maxage + "-got-" + scf.getMaxAge());
--- a/src/com/sun/ts/tests/servlet/api/jakarta_servlet_http/sessioncookieconfig/TestListener.java
+++ b/src/com/sun/ts/tests/servlet/api/jakarta_servlet_http/sessioncookieconfig/TestListener.java
@@ -80,8 +80,8 @@ public class TestListener implements ServletContextListener {
       scf.setComment(comment.toString());
     }
 
-    if (scf.getName() != null) {
-      comment.append("|getName-FAILED-expecting-null-got-" + scf.getName());
+    if (scf.getName() != null && !"JSESSIONID".equals(scf.getName()) {
+      comment.append("|getName-FAILED-expecting-null-or-JSESSIONID-got-" + scf.getName());
       scf.setComment(comment.toString());
     }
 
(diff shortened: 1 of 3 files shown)


scottmarlow:
Pending feedback from the Servlet leads, I could build a new Platform TCK tonight that excludes this test if they approve on Nov 1st in time.  I did commit to releasing the Platform TCK today, which I will do tonight.  

We are already planning on releasing a 5.0.1 Standalone Servlet TCK after EE 9 goes final.  We could also release a Platform TCK 9.0.1 as well then for this issue.

We are currently passing these tests with the Platform TCK + WildFly:
```
[javatest.batch] PASSED........com/sun/ts/tests/servlet/api/jakarta_servlet_http/sessioncookieconfig/URLClient.java#constructortest1
[javatest.batch] PASSED........com/sun/ts/tests/servlet/api/jakarta_servlet_http/sessioncookieconfig/URLClient.java#setCommentTest
[javatest.batch] PASSED........com/sun/ts/tests/servlet/api/jakarta_servlet_http/sessioncookieconfig/URLClient.java#setDomainTest
[javatest.batch] PASSED........com/sun/ts/tests/servlet/api/jakarta_servlet_http/sessioncookieconfig/URLClient.java#setHttpOnlyTest
[javatest.batch] PASSED........com/sun/ts/tests/servlet/api/jakarta_servlet_http/sessioncookieconfig/URLClient.java#setMaxAgeTest
[javatest.batch] PASSED........com/sun/ts/tests/servlet/api/jakarta_servlet_http/sessioncookieconfig/URLClient.java#setNameTest
[javatest.batch] PASSED........com/sun/ts/tests/servlet/api/jakarta_servlet_http/sessioncookieconfig/URLClient.java#setPathTest
[javatest.batch] PASSED........com/sun/ts/tests/servlet/api/jakarta_servlet_http/sessioncookieconfig/URLClient.java#setSecureTest
```

stuartwdouglas (approved):

scottmarlow:
Thanks Stuart for approving.  

@Thihup could you please provide the exact tests that you are seeing the failures with?  Is it just com/sun/ts/tests/servlet/api/jakarta_servlet_http/sessioncookieconfig/URLClient.java#setNameTest?

I need to move forward with promoting the Platform TCK (to unblock the remaining SPEC API ballots that are blocked on having a final Platform TCK).  But, it would be good to have the exact test names that you are seeing failures with that will be addressed by the pull request.

Thihup:
@scottmarlow 
Sorry, I didn't make myself clear.
Right now all the container implementations are passing the constructortest1 because it actually doesn't check for failures.
With this change, some containers will fail the constructortest1 because some of them return the string "JSESSIONID" as the result of the _SessionCookieConfig.getName()_ instead of being a null if the _SessionCookieConfig.setName()_ wasn't call  (as per the Javadoc [1]).

[1] https://jakarta.ee/specifications/servlet/5.0/apidocs/jakarta/servlet/SessionCookieConfig.html#getName()

stuartwdouglas:
I wonder if we should change the spec here. This behaviour seems really unintuitive. 

@gregw might also have an opinion here.

markt-asf:
My understanding is that `SessionCookieConfig.getName()` has to return `null` in this case so the caller can distinguish between no application specified name and explicit application configuration of the spec defined default name. This is because the priority order for determining the session cookie name is:

- application configured
- container configured default
- spec defined default

If `SessionCookieConfig.getName()` returns `JSESSIONID` when no name had been explicitly set then any container provided default would never be used.

We could change the API so that `SessionCookieConfig.getName()` implementations are required to implement the above priority order and then return the name used. However, that would result in asymmetric getters and setters - a scenario that has its own issues. Maybe there needs to be a new `getNameUsed()` method.

gregw:
I see it differently to @markt-asf as the javadoc of `setName` says the default is `JSESSION_ID`:
```java
    /**
     * Sets the name that will be assigned to any session tracking cookies created on behalf of the application represented
     * by the <tt>ServletContext</tt> from which this <tt>SessionCookieConfig</tt> was acquired.
     *
     * <p>
     * NOTE: Changing the name of session tracking cookies may break other tiers (for example, a load balancing frontend)
     * that assume the cookie name to be equal to the default <tt>JSESSIONID</tt>, and therefore should only be done
     * cautiously.
     *
     * @param name the cookie name to use
     *
     * @throws IllegalStateException if the <tt>ServletContext</tt> from which this <tt>SessionCookieConfig</tt> was
     * acquired has already been initialized
     */
    public void setName(String name);
```

Note also that the user never creates an instance of this interface, it is the container that does so.  Thus the container already has the ability to override with a container default before returning it to the application, which may then override.

So I think it is OK that containers can preset this (however they like) and not have to convert a null value back into either a container or spec default.... although we probably should then say what should happen if `setName(null)` is called?

tl;dr;  I do not think that `getName` should ever return null, unless that means do not use cookies.

markt-asf:
@gregw I think you are reading more than was intended into the use of the word default there. The Javadoc for `getName()` explicitly states it should return `null` if `setName()` has not been called.
I am concerned that changing the behaviour of `getName()` risks changing the session cookie name used by applications and thereby creating the sorts of issues described in the Javadoc for `setName()`.

gregw:
@markt-asf but the same javadoc for `getName` also says: `By default, JSESSIONID will be used as the cookie name`.
Moreover, when it says `null if {@link #setName} was never called`, it does not specify by whom the call was made.  Can the container call setName itself and thus satisfy the javadoc contract ?

I think the javadoc is ambiguous and at this stage it is going to be difficult to pick one way or another.     So perhaps we need to expect that getName() can return either null or the default value, but we need to define what null means - no cookie or default cookie?

standby and I'll propose some new wording....

markt-asf:
Yes. Tomcat returns `null` in that case. Setting the cookie name to `null` might make sense if the caller wants to clear a previously set value and resort to container (or spec) default. That seems like an unlikely use case now but may become more valid if we move in the direction of allowing more changes of this nature at runtime (as discussed elsewhere).

gregw:
Actually, I can't think of how to improve the wording whilst still allowing null to be returned.  If we want to not allow the name to be null, then we can tighten up the wording, but if null is a valid return then I think the ambiguity that exists in the javadoc actually matches reality.

Perhaps we could be stateful... allow a null return until the context has been initialised and then return the default... but that's ugly.

The more I think about this, the more allowing a null return is wrong/ugly.  It just means that in many places we are going to have to have code that checks for null and sets the default (eg instead of `scc.getName()` everybody has to use something like `scc.getName()==null?"JSESSIONID":scc.getName()`.     Encouraging that kind of code duplication is just silly and error prone.

So I'm very much for changing the spec to avoid allowing a null to be returned or to be set.

No matter what we do, I think jetty will always follow the spec document when it says:
```
 The standard name of the session tracking cookie must be `JSESSIONID`.
Containers may allow the name of the session tracking cookie to be
customized through container specific configuration.
```
by having out container specific configuration to always call `setName("JSESSIONID")` and thus never return null.

But for this PR, I think the test should be changed to allows either a null or "JSESSIONID" return.   I'll add review to that effect.

src/com/sun/ts/tests/servlet/api/jakarta_servlet_http/sessioncookieconfig/TestListener.java
```diff
@@ -47,48 +47,48 @@ public void contextInitialized(ServletContextEvent sce) {
...
     }
 
     if (scf.getMaxAge() != maxage) {
       comment.append(
-          "|getMaxAgeFAILED-expecting-" + maxage + "-got-" + scf.getMaxAge());
+          "|getMaxAge-FAILED-expecting-" + maxage + "-got-" + scf.getMaxAge());
       scf.setComment(comment.toString());
     }
 
     if (scf.getName() != null) {
```
gregw, line 83: We should allow for a container to a pre configured the name by calling `setName("JSESSIONID")`
```suggestion
    if (scf.getName() != null && !"JSESSIONID".equals(scf.getName()) {
```

gregw (changes requested):

gregw:
Is a empty string passed to `setName` a request for the default or an IAE?  What about illegal cookie names?

Thihup:
Similarly to the getName(), there are the getMaxAge() and getPath() that should return a specific default value (-1 for the max age and the null for the path).
Should I add both checks to this PR?

janbartel:
I don't see the problem with getMaxAge() the javadoc seems consistent - always return -1 unless setMaxAge() was called? But the javadoc for getPath/setPath is as messed up as getName/setName.

Thihup:
I've run the same test in some containers and these were the results of the default values for _getName_, _getMaxAge_ and _getPath_

| Container |getName() | getMaxAge() | getPath() |
|------|-------|------|-----|
| Payara | null | -1 | null |
| Tomcat | null | -1 | null |
| Jetty | JSESSIONID | -1 | null |
| Undertow | JSESSIONID | 0 | /servlet_pluh_sessioncookieconfig_web |
| Liberty | JSESSIONID | -1 | / |

scottmarlow:
CC @kwsutter @fl4via as per https://github.com/eclipse-ee4j/jakartaee-tck/pull/567#issuecomment-720516207, it looks like only WildFly + OpenLiberty will fail the SessionCookieConfig constructor test for Jakarta EE 9.1.  

There shouldn't be any new TCK test failures from changing getName() to be checked for either JSESSIONID or null.  

I think that Undertow will need a change to ensure that getMaxAge() returns -1.  

I think that both Undertow + OpenLiberty will need a change to ensure that getPath() returns null.

The current plan is to merge this pr in a few days.

Thihup:
Should I add a test for ensuring the default getPath is null?
I haven't added it yet because I'd like to know if returning null is the correct behavior.

scottmarlow:
> Should I add a test for ensuring the default getPath is null?
> I haven't added it yet because I'd like to know if returning null is the correct behavior.

No, I must of misunderstood.  Thanks for clarifying.  I will update my comment now. :)

gurunrao:
@Thihup - Any open comments pending for implementation?

arjantijms:
Just a ping @Thihup @gregw @markt-asf any updates here?

gurunrao:
@scottmarlow @Thihup @markt-asf @gregw - For the ^ PR to be merged and included with Jakarta EE 10 and servlet TCK, we have less time left.

markt-asf:
Looking at this now. I have the conflicts fixed locally. That will probably result in a new PR once I've confirmed the Servlet tests still pass for me.

gurunrao:
PR https://github.com/eclipse-ee4j/jakartaee-tck/pull/934 has been opened to address same.

PR #588, opened 2020-12-16
[587] Use `isInstance` rather than `==`

### LINKED ISSUE #587: JAX-RS: Provider class identity check should be instanceof check
The test in https://github.com/eclipse-ee4j/jakartaee-tck/blob/c40a006ac3c4d393eb2b5e8ca8d2244847d521fe/src/com/sun/ts/tests/jaxrs/ee/rs/ext/providers/ProvidersServlet.java#L103 checks that the `AnyExceptionExceptionMapper` provider class is properly registered and that the `Providers` implementation returns an instance of that exception mapper provider class.  The intent of the test is good, but the implementation checks whether the provider class returned from `Providers` is the exact same class as the one registered.  

This doesn't work in cases where CDI is used. Section 11.2.3 of the spec says:
> In a product that supports CDI, implementations MUST support the use of CDI-style Beans as root resource classes, providers and Application subclasses. Providers and Application subclasses MUST be singletons or use application scope.

Most CDI-style beans (and particularly application scoped beans) would return a proxy of the actual implementing class, which would fail an `==` check but would pass an `instanceof` check.  I suggest that the TCK test case be modified to:
```
if (em != null && AnyExceptionExceptionMapper.class.isInstance(em))
```

I'll submit a PR with this proposed change shortly.

andymc12:
Fixes #587

### DIFF
--- a/src/com/sun/ts/tests/jaxrs/ee/rs/ext/providers/ProvidersServlet.java
+++ b/src/com/sun/ts/tests/jaxrs/ee/rs/ext/providers/ProvidersServlet.java
@@ -104,7 +104,7 @@ public class ProvidersServlet extends ApplicationServlet {
     ExceptionMapper<RuntimeException> em = providers
         .getExceptionMapper(RuntimeException.class);
     Status status = Status.NOT_ACCEPTABLE;
-    if (em != null && em.getClass() == AnyExceptionExceptionMapper.class)
+    if (em != null && AnyExceptionExceptionMapper.class.isInstance(em))
       status = Status.OK;
     // This serverError() is to get ResponseBuilder with status != OK
     return Response.serverError().status(status).build();


jansupol (approved):

gurunrao (approved):
LGTM

RohitKumarJain (approved):
LGTM!!!

andymc12:
@scottmarlow @spericas Happy New Year! 

I don't have merge permission in this repo.  If you are ok with these changes, would you mind merging for me?

Thanks, Andy

scottmarlow:
> @scottmarlow @spericas Happy New Year!

Happy New Year!

> 
> I don't have merge permission in this repo. If you are ok with these changes, would you mind merging for me?

It is merged for the Jakarta EE 9.1 release, however, the best that we could do for Jakarta EE 9.0, is raise a TCK challenge and exclude the related test.

> 
> Thanks, Andy

andymc12:
Thanks for the merge, Scott. For 9.0, we'll plan to submit a challenge.

Issue #604, opened 2021-01-27
Remove org.omg.CORBA.ORB references from TCK tests


scottmarlow:
We should look for tests that reference a Corba ORB and move the Corba using tests to a separate test folder so that they may be made optional for Jakarta EE 9.1.  This is similar to what we did for JAXB using tests in Jakarta EE 9.

I'm thinking that we should enable the Corba ORB using tests by default.

@RohitKumarJain  Is this something that you could help with?

Note that @dblevins is going to remove the PortableRemoteObject#narrow calls from TCK tests via a separate pull request.

RohitKumarJain:
@scottmarlow , sure I'll start looking into this.

scottmarlow:
From our call this morning, a number of possible ideas were mentioned for dealing with the nested set of calls into the test bean that uses Corba.  Some observations from the call:

1. We don't know of a way to reference keyword properties in the actual test code so it would be hard to update the existing tests to directly check the `keywords` system property as that might not be set in the JVM process running the test.
2. The description for this issue suggests that we clone the Corba using test code into a new package, but that is more complicated for such a deep chain of calling code classes such as is shown below.  
3. One option is we could clone the actual test code into a corba package.  For example, clone com.sun.ts.tests.ejb30.misc.metadataComplete.warejb.TestServlet into com.sun.ts.tests.ejb30.misc.metadataComplete.warejbcorba.TestServlet and move the TestServlet.sameInterceptor3Times() into the warejbcorba.TestServlet class.
4. Do the same for all of the other test methods that depend on Corba.
5. If we come across test methods that are testing both Corba and not-Corba, we would need to ensure that both variations (Corba + non-Corba) are updated to offset the removal of some test logic.

Below is the call tree into com.sun.ts.tests.ejb30.bb.session.stateless.basic.RemoteCalculatorBean4Super#remoteAdd which depends on the injected ORB instance in various way:

> RemoteCalculatorBean4Super.remoteAdd(int, int)  (com.sun.ts.tests.ejb30.bb.session.stateless.basic)
> BaseLocalRemoteCalculator.remoteAdd(int, int)  (com.sun.ts.tests.ejb30.common.calc)
> TestServlet.sameInterceptor3Times(HttpServletRequest, HttpServletResponse)  (com.sun.ts.tests.ejb30.misc.metadataComplete.warejb)
[...]

PR #620, opened 2021-02-18
Make Corba ORB referencing tests optional by adding 'ExcludeCorba' system property check.  

scottmarlow:

---
Make Corba ORB referencing tests optional by adding 'ExcludeCorba' system property check.  If 'ExcludeCorba' is set to true, then tests should not expect to use Corba.
---

**Fixes Issue**
https://github.com/eclipse-ee4j/jakartaee-tck/issues/604

**Describe the change**
We cannot make reasonable use of `keywords` as we need to prevent the use of Corba within some tests without disabling the non-Corba aspects of those tests.  We normally would clone the specific part of tests that use the particular technology (Corba in this case) and have a new set of tests that are specific to that technology.  IMO, that doesn't seem to be a good match for the ejb30 tests as we would have to duplicate too much.

**Additional context**
We discussed removing the use of javax.rmi.PortableRemoteObject for which an additional pull request is still expected.

CC @alwin-joseph @anajosep @arjantijms @cesarhernandezgt @dblevins @m0mus @edbratt @gurunrao @jansupol @jgallimore @kazumura @kwsutter @LanceAndersen @bhatpmk @RohitKumarJain @shighbar @gthoman

### DIFF
--- a/src/com/sun/ts/tests/ejb30/common/system/Exclude.java
+++ b/src/com/sun/ts/tests/ejb30/common/system/Exclude.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright (c) 2021 Oracle and/or its affiliates. All rights reserved.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0, which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception, which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ */
+
+package com.sun.ts.tests.ejb30.common.system;
+
+/**
+ * A way for tests to check if an optional technology is excluded.  
+ * This is only for special technologies that cannot be reasonably 
+ * represented by a keyword property.  
+ *
+ * @author Scott Marlow
+ */
+public class Exclude {
+
+    /**
+     * @return true if Corba is excluded from TCK testing.
+     */
+    public static boolean ignoreCorba() {
+        return Boolean.getBoolean("ExcludeCorba");
+      }
+}
--- a/src/com/sun/ts/tests/ejb30/bb/session/stateless/annotation/envsharing/build.xml
+++ b/src/com/sun/ts/tests/ejb30/bb/session/stateless/annotation/envsharing/build.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0"?>
 <!--
 
-    Copyright (c) 2018 Oracle and/or its affiliates. All rights reserved.
+    Copyright (c) 2018, 2021 Oracle and/or its affiliates. All rights reserved.
 
     This program and the accompanying materials are made available under the
     terms of the Eclipse Public License v. 2.0, which is available at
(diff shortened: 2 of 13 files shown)


scottmarlow:
Will start a build + test of this change.  I'll comment here again after that completes.

src/com/sun/ts/tests/ejb30/common/system/Exclude.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+public class Exclude {
+
+    /**
+     * @return true if Corba is excluded from TCK testing.
+     */
+    public static boolean ignoreCorba() {
+        return Boolean.getBoolean("ExcludeCorba");
+      }
+}
```
RohitKumarJain, line 34: looks good... maybe we need to update Copyright to Copyright (c) 2021

RohitKumarJain (approved):
pretty good idea... LGTM !!!

src/com/sun/ts/tests/ejb30/common/system/Exclude.java
scottmarlow: Good catch!  I Fixed the date.

scottmarlow:
https://ci.eclipse.org/jakartaee-tck/job/jakartaee-tck-scottmarlow/job/corba_ejb/5/ is testing this change with GlassFish.  Since GlassFish does include a Corba implementation.

After we see a successful test run, I will request that a different implementation that doesn't include a Corba implementation tests this change (with the `ExcludeCorba` system property defined on the test + app server JVM processes.)  I will also share a link to the Jakarta EE Platform 9.1 TCK zip that contains this change.  I am hoping that we do not see any ClassNotFoundException on Corba classes with such testing but if we do, more test changes can be made.

scottmarlow:
https://ci.eclipse.org/jakartaee-tck/job/jakartaee-tck-scottmarlow/job/corba_ejb/5/ passed with GlassFish 6.0 on Java SE 8.

The Platform TCK zip is available via https://ci.eclipse.org/jakartaee-tck/job/jakartaee-tck-scottmarlow/job/corba_ejb/5/artifact/jakartaeetck-bundles/jakartaeetck-9.1.0.zip for testing

scottmarlow:
Good news, our testing with GlassFish 6.0 (on Java 8) passed with these changes!  There were two test runs that passed all of the ejb30 tests.  One test run was mentioned in the previous comment and the other one was with a [customized branch](https://github.com/scottmarlow/jakartaee-tck/tree/corba_ejb_default_no_corba) for which `com.sun.ts.tests.ejb30.common.system.Exclude#ignoreCorba` always returns true (for easier verification without needing to set additional system properties).  

https://ci.eclipse.org/jakartaee-tck/job/jakartaee-tck-scottmarlow/job/corba_ejb_default_no_corba/1/artifact/jakartaeetck-bundles/jakartaeetck-9.1.0.zip was built from the https://github.com/scottmarlow/jakartaee-tck/tree/corba_ejb_default_no_corba branch is available for others to test with.  

@dblevins would it makes sense for the PortableRemoteObject.narrow() calls to be removed still?  Or should we do somethinhg different, like only call PortableRemoteObject#narrow()  if the ExcludeCorba system property is false?

dblevins:
My concern with this approach is that it won't work for vendors that don't support CORBA because the `org.omg.CORBA.ORB` and similar classes are being imported.  The class likely won't load.

scottmarlow:
> My concern with this approach is that it won't work for vendors that don't support CORBA because the `org.omg.CORBA.ORB` and similar classes are being imported, but won't be on the classpath. The test itself likely won't load. The system property, which is logic in the class, is unlikely to even be reached.

I will try some tests with a JVM that eagerly loads classes.  I have an alternative change in mind to explore for the tests that initialise an ORB but only call ORB.toString().

scottmarlow:
Plan C is reviewing the use of Corba ORB and if it truly is only used for the ORB.toString() call, that isn't testing more than that orb.init can work which could be moved to a separate test that just verifies that orb.init works and could be excluded via a `!corba` keyword.

hussainnm:
Naïve suggestion here, shouldn't adding `org.glassfish.corba:glassfish-corba-omgapi:4.2.3` to the classpath get you the required classes to run with JDK11

scottmarlow:
> Naïve suggestion here, shouldn't adding `org.glassfish.corba:glassfish-corba-omgapi:4.2.3` to the classpath get you the required classes to run with JDK11

Thanks but IMO the Jakarta EE server implementation needs to integrate with `org.glassfish.corba:glassfish-corba-omgapi:4.2.3` also.  So it is probably not just being able to resolve the Corba API classes.

scottmarlow:
I pushed some changes to a new branch https://github.com/scottmarlow/jakartaee-tck/tree/corba_ejb_removeit.  I'll do some initially testing and comment again when the changes are passing with GlassFish.

scottmarlow:
I'm closing this pr and will open a new pr based on https://github.com/scottmarlow/jakartaee-tck/tree/corba_ejb_removeit

PR #626, opened 2021-02-26
config fixes for jsf, jpa tcks in jdk11

alwin-joseph:
**Fixes Issue**
Fixes config issue in jsf & jpa standlaone TCKs when run in JDK11.

- jpa tck : ts.jte.jdk11 is missing from the bundle
 - jsf tck : config was starting glassfish with default jdk8 before setting java home to jdk11

**Describe the change**
jpa tck - added ts.jte.jdk11  to bundle
jsf tck - reordered setting java home before asadmin startdomain call  

This will pass the jsf tests, but more issues in jpa tck run to be fixed.

### DIFF
--- a/docker/jsftck.sh
+++ b/docker/jsftck.sh
@@ -1,6 +1,6 @@
 #!/bin/bash -xe
 
-# Copyright (c) 2018, 2020 Oracle and/or its affiliates. All rights reserved.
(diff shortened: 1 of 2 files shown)


gurunrao (approved):
LGTM

PR #654, opened 2021-03-12
#627 Class Not Found: JDBC driver org.apache.derby.jdbc.EmbeddedDriver could not be loaded

hussainnm:
Fix derbyEmbedded.classpath to include derbyshared.jar and derbytools.jar

### DIFF
--- a/install/jakartaee/bin/ts.jte.jdk11
+++ b/install/jakartaee/bin/ts.jte.jdk11
@@ -555,7 +555,7 @@ derbyEmbedded.user=cts1
 derbyEmbedded.passwd=cts1
 derbyEmbedded.url=jdbc:derby:${derbyEmbedded.dbName};create=true
 derbyEmbedded.driver=org.apache.derby.jdbc.EmbeddedDriver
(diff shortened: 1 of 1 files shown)


scottmarlow:
Seems likely that this is needed since org.apache.derby.jdbc.EmbeddedDriver is in the derbytools.jar (used to be in derby.jar)

scottmarlow (approved):

scottmarlow:
+1 for also adding derbyshared.jar

hussainnm:
Do I need to make the same changes to ts.jte too?

LanceAndersen (approved):

LanceAndersen:
> Do I need to make the same changes in ts.jte too?

derbyshared.jar is new starting with 10.15.1.3 which was built with OpenJDK 64-Bit Server VM 18.9 (build 11+28, mixed mode).  So if ts.jte is still intended to be used with JDK 8, I would only add a comment to add the jar if 10.15.1.3 is used

hussainnm:
Understood. No changes are required for ts.jte.

PR #656, opened 2021-03-13
#627 Class Not Found: JDBC driver org.apache.derby.jdbc.ClientDriver could not be loaded

hussainnm:
Fix derby.classes.ri to include derbyshared.jar and derbytools.jar


scottmarlow (approved):

scottmarlow:
Thank you @hussainnm for the fix!  Very much appreciated!

PR #664, opened 2021-03-15
Updates to JSTL bucket for Java 11 Locale

brideck:
Locale related changes to goldenfiles and JSPs to be included in the TCK once support for Java 8 is no longer needed.

Signed-off-by: Brian Decker <bmdecker@us.ibm.com>


pnicolucci (commented):
Overall these changes make sense and look good. My only comment would be we should probably update the copyrights where appropriate to include: `Copyright (c) 2021-2021 Contributors to the Eclipse Foundation`

What do you think?

brideck:
> 
> 
> > > Overall these changes make sense and look good. My only comment would be we should probably update the copyrights where appropriate to include: `Copyright (c) 2021-2021 Contributors to the Eclipse Foundation`
> > > What do you think?
> > 
> > 
> > Updating the copyright date in the updated files sounds right. For example, change from `Copyright (c) 2003, 2020 Oracle and/or its affiliates. All rights reserved.` to `Copyright (c) 2003, 2021 Oracle and/or its affiliates. All rights reserved.`
> 
> Hey @scottmarlow @brideck wouldn't we want to leave the Oracle copyright alone and add the general Eclipse copyright?

Not saying we shouldn't, but I interestingly only see an Eclipse copyright statement like the one Paul proposed in a few files in the TCK.  @scottmarlow, are we wanting to transition to that statement as we make changes to TCK files going forward?

PR #675, opened 2021-03-19
Restore copyright for java files not updated this year

dblevins:
Done by comparing the file contents from the `9.0.0` tag against current contents. For any files that had the exact same contents, sans copyright, the old file was restored as-is.

This is so we can see what has actually changed between the EE 9.0 and current 9.1 branches.  480 files were updated in copyright only.  We should either update all files or no files.


scottmarlow (approved):

scottmarlow:
I had removed most of the unused imports and wildcard imports as well but we ended up reverting the changes for files addressed by your pr.  I had thought we should keep the `2021` copyright update but I think this change is better since the (reverted) file contents are the same as they were in 2020.

PR #703, opened 2021-04-06
Fix the referenced jar-file element to match the jar files created in the build.xml

jeanouii:
Signed-off-by: Jean-Louis Monteiro <jlmonteiro@tomitribe.com>

This PR fixes the <jar-file> element of the persistence.xml to match the jar names created by the build.xml

They currently don't match which results in an undefined behaviour.
See https://github.com/eclipse-ee4j/jpa-api/issues/299 for more details

Glassfish seems to warn and ignore, but it's not portable.
In Apache TomEE it fails for instance.


lukasj (approved):
agree, this change makes sense

gurunrao (approved):
LGTM

RohitKumarJain (approved):
LGTM

PR #708, opened 2021-04-08
Fix JRuby issue

tvallin:
Signed-off-by: tvallin <thibault.vallin@oracle.com>


lukasj:
@gurunrao this should help but needs to be tested by jenkins

PR #709, opened 2021-04-08
Remove duplicate property maven.deploy.skip in UG jacc/pom.xml

tvallin:
Signed-off-by: tvallin <thibault.vallin@oracle.com>


gurunrao:
@tvallin - mvn call - https://github.com/eclipse-ee4j/jakartaee-tck/blob/master/release/tools/jakartaee.xml#L148 is the root-cause of build failure.

gurunrao (approved):

PR #704, opened 2021-04-06
Use a transaction scoped EntityManager to set/get FlushMode

jeanouii:
Signed-off-by: Jean-Louis Monteiro <jlmonteiro@tomitribe.com>

This PR fixes the set/get test for FlushMode
In short the EntityManager is not bound to a Transaction so the set and get of the FlushMode happens on a different instance.

More details available on this issue https://github.com/eclipse-ee4j/jpa-api/issues/300

PR #721, opened 2021-04-15
Use a transaction scoped EntityManager to set/get a property

jeanouii:
Signed-off-by: Jean-Louis Monteiro jlmonteiro@tomitribe.com

Same bug as the one fixed with https://github.com/eclipse-ee4j/jakartaee-tck/pull/704
This PR fixes the set/get test for EntityManager property
In short the EntityManager is not bound to a Transaction so the set/get happens on a different instance.

### DIFF
--- a/src/com/sun/ts/tests/jpa/core/entityManager/Client.java
+++ b/src/com/sun/ts/tests/jpa/core/entityManager/Client.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2007, 2020 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2007, 2021 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
--- a/src/com/sun/ts/tests/jpa/core/entityManager/Client.java
+++ b/src/com/sun/ts/tests/jpa/core/entityManager/Client.java
@@ -34,6 +34,7 @@ import jakarta.persistence.EntityExistsException;
 import jakarta.persistence.EntityManager;
 import jakarta.persistence.EntityManagerFactory;
 import jakarta.persistence.EntityNotFoundException;
+import jakarta.persistence.EntityTransaction;
 import jakarta.persistence.LockModeType;
 import jakarta.persistence.ParameterMode;
 import jakarta.persistence.PersistenceException;
@@ -1276,6 +1277,8 @@ public class Client extends PMClientBase {
     boolean foundValue = false;
 
     try {
+      EntityTransaction t = getEntityTransaction();
+      t.begin();
       EntityManager em = getEntityManager();
       String expectedKey = "jakarta.persistence.cache.retrieveMode";
       CacheRetrieveMode expectedValue = CacheRetrieveMode.USE;
@@ -1319,6 +1322,15 @@ public class Client extends PMClientBase {
       }
     } catch (Exception e) {
       TestUtil.logErr("Unexpected exception occurred", e);
+
+    } finally {
+      try {
+        if (getEntityTransaction().isActive()) {
+          getEntityTransaction().rollback();
+        }
+      } catch (Exception fe) {
+        TestUtil.logErr("Unexpected exception rolling back TX:", fe);
+      }
     }
     if (!foundKey || !foundValue) {
       throw new Fault("setPropertyTest failed");


lukasj (commented):
I do not see different `em` instances being used by `set/get` here. If test fails, then it is clear that `em.get` did not found the property set earlier through `em.set` on the same instance of the entity manager and the implementation is wrong. Presence or absence of transaction should not be relevant here - the behaviour of `em.set/get` should be consistent. Ie `em.find` does not require active transaction and should allow setting a property/hint.

Since current test can be shortened to following JUnit equivalent:
```java
EntityManager em = getEntityManager();
String expectedKey = "jakarta.persistence.cache.retrieveMode";
CacheRetrieveMode expectedValue = CacheRetrieveMode.USE;
em.setProperty(expectedKey, expectedValue);
Assert.assertEquals(expectedValue, em.getProperties().get(expectedKey));
 ```
I do not think this test is wrong. Am I missing something?

**Comment by dblevins** [comment]
Here's the crux of the issue and any thoughts are very welcome.

 - The test uses a Container-managed Transaction-scoped Persistence Context (7.6.2), `PersistenceContextType.TRANSACTION`.  Per spec, the persistence context is tied to the transaction in progress  (paragraph 2).  When there is no transaction in progress there is no persistence context and any entities referenced are immediately detached (paragraph 4).  The most critical part of that is the definition that when there is no transaction, there is no persistence context.

 - The spec does define `PersistenceContextType.EXTENDED` persistence contexts, which are essentially the same as Container-managed Transaction-scoped Persistence Context but with the twist that the EntityManager reference can also be used outside of a transaction and have a persistence context.   This test is not using EXTENDED, so there is no persistence context of any kind in play for the EntityManager reference when the test uses it outside the scope of a transaction.

 - Section 3.9.2 defines `setProperty` as a way to set properties "specified at the level of the persistence context."  As defined by the spec via section 7.6.2, use of an EntityManager of type `PersistenceContextType.TRANSACTION` requires a transaction for there to be a persistence context.  Therefore use of `setProperty` on an EntityManager of type `PersistenceContextType.TRANSACTION` outside the scope of a transaction has no spec-defined behavior.

The test as written is testing unspecified vendor-specific behavior.  For it to be portable behavior it would need to start a transaction, which creates a spec-defined persistence context, and perform both the set and get in that same transaction.

While it's great that many implementations pass this test, the scope of the persistence contexts they're creating to pass it are undefined.  Some could scope it to the duration of stateless bean invocation, the duration of the HTTP request, the duration of the HTTP session.  For us, every invocation on the EntityManager reference outside a transaction will result in a persistence context that lives for that exact invocation on the EntityManager and is immediately discarded.  We do this intentionally so that users don't get rewarding behavior doing something not spec defined, and therefore not guaranteed portable.  We'd rather they get confused enough to learn to do it the proper way.

The only spec-defined way a Container-managed Transaction-scoped Persistence Context that lives for any duration is in a transaction.

The test should either be fixed to use a transaction or excluded as it tests something not backed by the spec.

**Comment by scottmarlow** [comment]
@dblevins @jeanouii can you please attach the .jtr test output for the failure you are seeing without the code change (look for .jtr files(s) with the failure under `jakartaeetck/bin/JTwork/com/sun/ts/tests/jpa/core/entityManager/` folder).

**Comment by dblevins** [comment]
I'm not entirely sure that's relevant, but here it is.  The ".txt" extension is added so Github will accept it.

[Client_setPropertyTest_from_stateless3.jtr.txt](https://github.com/eclipse-ee4j/jakartaee-tck/files/6321794/Client_setPropertyTest_from_stateless3.jtr.txt)

**Comment by jeanouii** [comment]
Thanks @dblevins 

This is the same situation as the one we already fixed
https://github.com/eclipse-ee4j/jakartaee-tck/pull/704

**Comment by scottmarlow** [comment]
> Thanks @dblevins
> 
> This is the same situation as the one we already fixed
> #704

Was the #704 failure also with stateless3 as is shown in the attached `Client_setPropertyTest_from_stateless3.jtr.txt`?

Description of the stateless3 vehicle quoting from https://github.com/eclipse-ee4j/jakartaee-tck/blob/master/user_guides/jakartaee/src/main/jbake/content/config.adoc:
```
`stateless3`: Bean-managed stateless session bean using JNDI to lookup
a Jakarta Transactions `EntityManager`; uses `UserTransaction` methods for transaction
demarcation
```

I think the pending question is whether the test needs to start the EntityTransaction itself or if it should already be started for each test that runs in the stateless3 vehicle?

lukasj:
I think the missing and likely critical piece of information here was that the test runs within the context of *Bean Managed Transactions* (see [the vehicle](https://github.com/eclipse-ee4j/jakartaee-tck/blob/master/src/com/sun/ts/tests/common/vehicle/stateless3/Stateless3VehicleBean.java)), so what we have is _a JTA application-managed entity manager created outside the scope of the current JTA transaction_.

lukasj (approved):
@dblevins makes sense. Would be great if all PRs contain such comprehensive description

dblevins:
> I think the pending question is whether the test needs to start the EntityTransaction itself or if it should already be started for each test that runs in the stateless3 vehicle?

@scottmarlow I'm not too sure.  At some point, we may want to investigate to see how that vehicle is used in other tests.  There are definitely spec requirements that require the `EntityManager` to throw specific exceptions like `TransactionRequiredException` for Container-managed Transaction-scoped Persistence Context's used outside a transaction, such as `EntityManger.refresh`.

PR #760, opened 2021-10-08
reduce container/vm memory + cpu settings, reduce -Xmx max heap size

scottmarlow:

Use one CPU core per running TCK test and also reduce memory usage as well.  I expect there could be some EJB30 tests settings may need to be adjusted to run fewer tests per test group due to lower JVM max memory sizing.


scottmarlow:
I'm closing this so we don't see it in the pull request list.  Others can work on it or perhaps I will again at some point.

PR #800, opened 2022-01-18
update docker scripts for jdk17.

gurunrao:
update docker scripts for jdk17.
Signed-off-by: gurunandan.rao@oracle.com <gurunandan.rao@oracle.com>


alwin-joseph (approved):
LGTM.

docker/build_jakartaeetck.sh
```diff
@@ -18,7 +18,6 @@ if [ -z "$ANT_HOME" ]; then
   export ANT_HOME=/usr/share/ant/
 fi
```
scottmarlow, line 20: @gurunrao 

I think that we should always build with `JDK11` for EE 10.  With this pull request, we should only be changing the running of TCK tests to be able to use either `JDK11` or `JDK17`.

docker/build_standalone-tcks.sh
```diff
@@ -27,7 +27,6 @@ if [ -z "$JAKARTA_JARS" ]; then
   export JAKARTA_JARS=$BASEDIR
 fi
 
-export JAVA_HOME=${JDK11_HOME}
```
scottmarlow, line 30: I think that we should always build with JDK11 for EE 10. With this pull request, we should only be changing the running of TCK tests to be able to use either JDK11 or JDK17.

docker/build_jakartaeetck.sh
gurunrao: With update to docker image "jakartaee/cts-base:0.3" default JAVA_HOME is set to "/opt/jdk-11.0.2"(replacing jdk8 java_home), hence we need not set JAVA_HOME for build.

docker/build_standalone-tcks.sh
gurunrao: With update to docker image "jakartaee/cts-base:0.3" default JAVA_HOME is set to "/opt/jdk-11.0.2"(replacing jdk8 java_home), hence we need not set JAVA_HOME at build script(have removed redundant JAVA_HOME export)

scottmarlow (approved):

Jenkinsfile
```diff
@@ -133,7 +133,7 @@ spec:
       limits:
         memory: "3Gi"
   - name: jakartaeetck-ci
-    image: jakartaee/cts-base:0.2
+    image: jakartaee/cts-base:0.3
```
pzygielo, line 136: Could https://github.com/eclipse-ee4j/jakartaee-tck/tree/master/docker-images/cts-base be updated with Dockerfile used to produce 0.3 + entry in README, please?

alwin-joseph: Created https://github.com/eclipse-ee4j/jakartaee-tck/pull/824. Thanks!

PR #841, opened 2022-02-14
Remove the api version mentioned in the signature files

alwin-joseph:
**Fixes Issue**
No Issues created.

**Describe the change**

This change removes the occurrence of API versions in the signature files at src/com/sun/ts/tests/signaturetest/signature-repository. This will remove the ability to have multiple versions of signature files for an API as there will be only one signature file for each API. This change is made for the convenience of not maintaining the versions in the signature files. It is not anticipated that there will be requirement to keep multiple signature files for a particular API at the moment. If at all this becomes a necessity this change can be revered too.

Update:
Latest commit also removed the "_se11" suffix in sigtest map file & sigtest package list file names  and the values(versions) corresponding to the APIs within the sig map files. 

There are more changes that can be considered along with this : 
- Remove the mapping file entirely, as the only reason for their existence after this change is to pick the right version of the APIs to run the signature tests. In this case we would need to provide the API version via some other option maybe via ts.jte for the test runs.


alwin-joseph:
Update:
Latest commit also removed the "_se11" suffix in sigtest map file & sigtest package list file names  and the values(versions) corresponding to the APIs within the sig map files.

install/jakartaee/bin/sig-test.map
```diff
@@ -26,16 +26,12 @@
 #
 # The resultant signature file name has a name that complies
 # with the following format:
-# <package>.sig_<technology_version><JavaSE_version_these_work_with>
+# <package>.sig_<technology_version>
```
scottmarlow, line 29: Should the `<technology_version>` also be remove such that we end up signatures like `jakarta.annotation.sig`?

https://github.com/scottmarlow/jakartaee-tck/commit/9b7cd83b718dcbf3d565811468a674533574e5c5 shows what I mean (your pull request is more complete but you didn't seem to remove the api version).

src/com/sun/ts/tests/signaturetest/SignatureTestDriver.java
```diff
@@ -498,23 +479,19 @@ protected Properties loadMapFile(String mapFile)
...
 
     String sigFile;
     if (repositoryDir.endsWith(File.separator)) {
-      sigFile = repositoryDir + baseName + SIG_FILE_EXT + SIG_FILE_VER_SEP
-          + version;
+      sigFile = repositoryDir + baseName + SIG_FILE_EXT;
     } else {
-      sigFile = repositoryDir + File.separator + baseName + SIG_FILE_EXT
-          + SIG_FILE_VER_SEP + version;
+      sigFile = repositoryDir + File.separator + baseName + SIG_FILE_EXT;
```
scottmarlow, line 494: :+1: as you did remove the api version here.

install/jakartaee/bin/sig-test.map
alwin-joseph: Corrected this comment and others.

scottmarlow (approved):

PR #843, opened 2022-02-14
Avoid data that could lead to ambiguous column label usage

beikov:
See https://github.com/eclipse-ee4j/jpa-api/issues/350 for further discussion

**Fixes Issue**
Specify the issue (link) that is solved with this pull request.

**Related Issue(s)**

**Describe the change**
A clear and concise description of the change.

**Additional context**

### DIFF
--- a/src/com/sun/ts/tests/jpa/core/annotations/nativequery/Client.java
+++ b/src/com/sun/ts/tests/jpa/core/annotations/nativequery/Client.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2007, 2020 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2007, 2022 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
@@ -76,9 +76,9 @@ public class Client extends PMClientBase {
       getEntityTransaction().begin();
 
       TestUtil.logTrace("Create Items");
-      final Item i1 = new Item(4, "WaterShoes");
-      final Item i2 = new Item(5, "FlipFlops");
-      final Item i3 = new Item(6, "Sandals");
+      final Item i1 = new Item(1, "WaterShoes");
+      final Item i2 = new Item(2, "FlipFlops");
+      final Item i3 = new Item(3, "Sandals");
 
       TestUtil.logTrace("Create Order1s");
       Order1 o4 = new Order1(4, 25.0D);
@@ -188,9 +188,9 @@ public class Client extends PMClientBase {
       getEntityTransaction().begin();
 
       TestUtil.logTrace("Create Items");
-      final Item i1 = new Item(7, "Loafers");
-      final Item i2 = new Item(8, "High Heels");
-      final Item i3 = new Item(9, "Socks");
+      final Item i1 = new Item(1, "Loafers");
+      final Item i2 = new Item(2, "High Heels");
+      final Item i3 = new Item(3, "Socks");
 
       TestUtil.logTrace("Create Orders");
 
@@ -303,9 +303,9 @@ public class Client extends PMClientBase {
       getEntityTransaction().begin();
 
       TestUtil.logTrace("Create Items");
-      final Item i1 = new Item(1, "Boots");
-      final Item i2 = new Item(2, "Sneakers");
-      final Item i3 = new Item(3, "Slippers");
+      final Item i1 = new Item(4, "Boots");
+      final Item i2 = new Item(5, "Sneakers");
+      final Item i3 = new Item(6, "Slippers");
 
       TestUtil.logTrace("Create Orders");
       Order1 o1 = new Order1(1, 25.0D);
@@ -410,9 +410,9 @@ public class Client extends PMClientBase {
       getEntityTransaction().begin();
 
       TestUtil.logTrace("Create Items");
-      final Item i1 = new Item(7, "Loafers");
-      final Item i2 = new Item(8, "High Heels");
-      final Item i3 = new Item(9, "Socks");
+      final Item i1 = new Item(1, "Loafers");
+      final Item i2 = new Item(2, "High Heels");
+      final Item i3 = new Item(3, "Socks");
 
       TestUtil.logTrace("Create Orders");
 
@@ -518,9 +518,9 @@ public class Client extends PMClientBase {
       getEntityTransaction().begin();
 
       TestUtil.logTrace("Create Items");
-      final Item i1 = new Item(7, "Loafers");
-      final Item i2 = new Item(8, "High Heels");
-      final Item i3 = new Item(9, "Socks");
+      final Item i1 = new Item(1, "Loafers");
+      final Item i2 = new Item(2, "High Heels");
+      final Item i3 = new Item(3, "Socks");
 
       TestUtil.logTrace("Create Orders");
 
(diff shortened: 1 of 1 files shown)


scottmarlow:
I see nothing wrong with the change myself.  We would like to merge this change soon if the test truly needs it but first please could I get some help with understanding if this change is needed even though SqlResultSetMappings is specified?  See more details below...

Should the the [@SqlResultSetMappings in Order1 have avoided the need for this change?](https://github.com/eclipse-ee4j/jakartaee-tck/blob/master/src/com/sun/ts/tests/jpa/core/annotations/nativequery/Order1.java#L37)

Or is [SqlResultSetMappings (click here for spec text)](https://github.com/eclipse-ee4j/jakartaee-tck/blob/master/src/com/sun/ts/tests/jpa/core/annotations/nativequery/Order1.java#L37) for something else?  

Note that SqlResultSetMappings is also mentioned in section [3.10.16.1. Returning Managed Entities from Native Queries](https://jakarta.ee/specifications/persistence/3.0/jakarta-persistence-spec-3.0.html#returning-managed-entities-from-native-queries) which was discussed in the issue but I didn't see `SqlResultSetMappings` mentioned in the issue and wonder if the `SqlResultSetMappings` should of helped.

beikov:
Changing the data will uncover that the test is "wrong", but for further information on that, see https://github.com/eclipse-ee4j/jpa-api/issues/350.

I tested this on EclipseLink and it will produce "wrong" results. The problem is that the test goes against the wording in 3.10.16.1, because the result label `id` appears twice in the result set metadata of the [test](https://github.com/eclipse-ee4j/jakartaee-tck/blob/da6a34066dd04a78e6b1656c923c9c5de97af182/src/com/sun/ts/tests/jpa/core/annotations/nativequery/Client.java#L95). To fix that, the query and the `SqlResultSetMapping` have to be adapted:

```java
@SqlResultSetMapping(name = "Order1ItemResults", entities = {
        @EntityResult(entityClass = com.sun.ts.tests.jpa.core.annotations.nativequery.Order1.class, fields = {
            @FieldResult(name = "id", column = "OID"),
            @FieldResult(name = "totalPrice", column = "OPRICE"),
            @FieldResult(name = "item", column = "OITEM") }) }),
        @EntityResult(entityClass = com.sun.ts.tests.jpa.core.annotations.nativequery.Item.class) }),
```

```java
q = getEntityManager().createNativeQuery(
          "Select o.\"ID\" OID, o.\"TOTALPRICE\" OPRICE, "
              + " o.\"FK1_FOR_ITEM\" OITEM, i.\"ID\", i.\"ITEMNAME\" from \"ORDER1\" o, \"ITEM\" i "
              + "WHERE (o.\"TOTALPRICE\" > 140) AND (o.\"FK1_FOR_ITEM\" = i.\"ID\")",
          "Order1ItemResults").getResultList();
```

In fact, `Order3ItemResults` and [`nativeQueryTest3`](https://github.com/eclipse-ee4j/jakartaee-tck/blob/da6a34066dd04a78e6b1656c923c9c5de97af182/src/com/sun/ts/tests/jpa/core/annotations/nativequery/Client.java#L322) do exactly that. So maybe `nativeQueryTest1` should just be changed to a negative test i.e. assert an exception is thrown?

scottmarlow:
> Changing the data will uncover that the test is "wrong", but for further information on that, see [eclipse-ee4j/jpa-api#350](https://github.com/eclipse-ee4j/jpa-api/issues/350).
> 
> I tested this on EclipseLink and it will produce "wrong" results. The problem is that the test goes against the wording in 3.10.16.1, because the result label `id` appears twice in the result set metadata of the [test](https://github.com/eclipse-ee4j/jakartaee-tck/blob/da6a34066dd04a78e6b1656c923c9c5de97af182/src/com/sun/ts/tests/jpa/core/annotations/nativequery/Client.java#L95). To fix that, the query and the `SqlResultSetMapping` have to be adapted:
> 
> ```java
> @SqlResultSetMapping(name = "Order1ItemResults", entities = {
>         @EntityResult(entityClass = com.sun.ts.tests.jpa.core.annotations.nativequery.Order1.class, fields = {
>             @FieldResult(name = "id", column = "OID"),
>             @FieldResult(name = "totalPrice", column = "OPRICE"),
>             @FieldResult(name = "item", column = "OITEM") }) }),
>         @EntityResult(entityClass = com.sun.ts.tests.jpa.core.annotations.nativequery.Item.class) }),
> ```
> 
> ```java
> q = getEntityManager().createNativeQuery(
>           "Select o.\"ID\" OID, o.\"TOTALPRICE\" OPRICE, "
>               + " o.\"FK1_FOR_ITEM\" OITEM, i.\"ID\", i.\"ITEMNAME\" from \"ORDER1\" o, \"ITEM\" i "
>               + "WHERE (o.\"TOTALPRICE\" > 140) AND (o.\"FK1_FOR_ITEM\" = i.\"ID\")",
>           "Order1ItemResults").getResultList();
> ```
> 

Thanks for explaining, the example makes sense!



> In fact, `Order3ItemResults` and [`nativeQueryTest3`](https://github.com/eclipse-ee4j/jakartaee-tck/blob/da6a34066dd04a78e6b1656c923c9c5de97af182/src/com/sun/ts/tests/jpa/core/annotations/nativequery/Client.java#L322) do exactly that. So maybe `nativeQueryTest1` should just be changed to a negative test i.e. assert an exception is thrown?

To explore that, the `nativeQueryTest1` test assertionIDs from 
[JavaDoc](https://github.com/eclipse-ee4j/jakartaee-tck/blob/master/install/jpa/docs/assertions/api/api-assertions-3_1.html) + [SPEC](https://github.com/eclipse-ee4j/jakartaee-tck/blob/master/install/jpa/docs/assertions/spec/spec-assertions-3_0.html) should be examined to understand what the test is supposed to validate exactly:

PERSISTENCE:SPEC:736
```
The SQL query facility is intended to provide support for those cases where it is necessary to use the native SQL of the target database in use (and/or where Jakarta Persistence QL cannot be used). Native SQL queries are not expected to be portable across databases. 
```

PERSISTENCE:SPEC:1009
```
The SqlResultSetMapping annotation is used to specify the mapping of the result of a native SQL query or stored procedure. 
```

PERSISTENCE:JAVADOC:199 (returned EntityResult[], for jakarta.persistence.SqlResultSetMapping.entities)
```
Specifies the result set mapping to entities.
```

PERSISTENCE:JAVADOC:64
```
The class of the result.
```

PERSISTENCE:JAVADOC:41
```
Create an instance of Query for executing a native SQL query.
```

Since the current Spec text does not make it clear that an exception is expected for the negative test case idea (unless I am missing something), I do not think we need the TCK test to validate that an exception is thrown for the current test.  

More background on test assertions in general are listed on https://github.com/eclipse-ee4j/jakartaee-tck/wiki/Guide-to-adding-Specification-Assertions

From my reading of the test change and the detail given about why the test needs to change, I think we should merge the change.  Will leave it open for another day to see if we get more feedback first.

Thanks for the help and contribution @beikov

beikov:
Note that this data change would simply show that the test is broken and would clearly show to implementers that their implementations (Hibernate, EclipseLink) just can't make this magic happen that the test seemed to expect.

IMO, the test should simply be disabled or changed in a way such that it behaves like `nativeQueryTest3`

scottmarlow:
> Note that this data change would simply show that the test is broken and would clearly show to implementers that their implementations (Hibernate, EclipseLink) just can't make this magic happen that the test seemed to expect.
> 
> IMO, the test should simply be disabled or changed in a way such that it behaves like `nativeQueryTest3`

Could you paste in here what  the updated `nativeQueryTest1` method should look like after it is modified to behave like `nativeQueryTest3`?  Or perhaps update this pull request.  That is if you have an idea in mind of how to make the change.

scottmarlow:
Does anyone plan on updating this pr to update `nativeQueryTest1` as mentioned above to address the challenge issue https://github.com/eclipse-ee4j/jpa-api/issues/350?  

I expect that https://github.com/eclipse-ee4j/jpa-api/issues/350 will be marked `accepted`/`invalid` on Friday March 4th, so that we should be able to take action on this pull request at that time or exclude the challenged `nativeQueryTest1` test.

beikov:
I can do that, I just need to know whether we want to adapt the test so that it works like `nativeQueryTest3` (which would make it useless I guess?), or if we want to change the test to assert an exception should be thrown.
I had the feeling that @lukasj should respond to my last comment on https://github.com/eclipse-ee4j/jpa-api/issues/350 before I start working on this, so that we all understand why that test is wrong.

beikov:
As discussed in https://github.com/eclipse-ee4j/jpa-api/issues/350 we will remove the `nativeQueryTest1`

PR #852, opened 2022-02-19
[xmlws40] update ug, install docs, removal of endorsed

lukasj:
**Describe the change**
* update xml ws tck user guide for XML WS 4.0, fix formating, minor issues, move the build to require SE 11 etc
* update release notes
* created assertion files for 4.0 from their 3.0 version (need to be updated separately/different PR(s))
* removed references to endorsed and resolver jar - both are invalid in context of XML WS 4.0 and SE 11
* fixed references to javax filenames, not sure they're in use, so just to let it be less confusing 


lukasj:
btw: what are those `ts.jte.tomcat.jdk1X` good for? My first impression was that the number means the JDK version - 16, 17 etc but thinking about it it's likely more about 1.6, 1.7 etc versions. If the latter is the right meaning, then these files should just go away (with javax->jakarta change and move to SE 11, I see no reason why different configs for different SE versions would be needed)

scottmarlow:
> btw: what are those `ts.jte.tomcat.jdk1X` good for? 

Not much as they all seem to be for older Java SE versions.

`jakartaee-tck/install/jaxws/bin/ts.jte.tomcat.jdk18` references Java SE 8+
`jakartaee-tck/install/jaxws/bin/ts.jte` references Java SE 6+
`jakartaee-tck/install/jaxws/bin/ts.jte.jdk19` references Java SE 6+
`jakartaee-tck/install/jaxws/bin/ts.jte.tomcat.jdk16` references Java SE 6+
`jakartaee-tck/install/jaxws/bin/ts.jte.tomcat.jdk17` references Java SE 6+
`jakartaee-tck/install/jaxws/bin/ts.jte.tomcat.moxy` references Java SE 6+

> My first impression was that the number means the JDK version - 16, 17 etc but thinking about it it's likely more about 1.6, 1.7 etc versions. If the latter is the right meaning, then these files should just go away (with javax->jakarta change and move to SE 11, I see no reason why different configs for different SE versions would be needed)

:+1: , I am guessing that `jakartaee-tck/install/jaxws/bin/ts.jte.tomcat.jdk18` could be kept and the older `tomcat` files removed except for `jakartaee-tck/install/jaxws/bin/ts.jte.jdk19` which IMO doesn't need the JDK version in the file name (so remove ts.jte and move `ts.jte.jdk19` to `ts.jte`

lukasj:
`ts.jte.jdk19 ` seems to be removed by different PR, so I removed `ts.jte.tomcat.jdk16` and `ts.jte.tomcat.jdk17` only

gurunrao (approved):

scottmarlow (approved):

PR #893, opened 2022-03-14
avoid OOM error while deployment in jsf tck

alwin-joseph:
**Fixes Issue**
The jsf tck tests are stuck during deployment due to OOM error as seen in glassfish server.logs

**Describe the change**
Doubles the maximum memory allocation for glassfish for jsf tck tests.


gurunrao (approved):

dmatej:
Note for myself: check if it is still required and if it is, find the cause and fix it.

PR #906, opened 2022-03-24
support jdk18, also allow enable/disable of security manager.

gurunrao:
Signed-off-by: gurunrao <gurunandan.rao@oracle.com>
Allow to disable/enable java security manager from jenkin jobs.


scottmarlow:
I think that this pull request is only changing GlassFish specific script files at the moment.  I would like to see the script files be generalized to work with any application server but that is not an EE 10 goal, just wanted to mention the idea.

Note my feedback entered yesterday in https://github.com/eclipse-ee4j/jakartaee-tck/pull/894 regarding making a change so that implementations like GlassFish continue to use the Java SecurityManager with Java SE 18.

scottmarlow:
This change looks good to me, I hope it gets reopened or shows up as a pull request.

PR #915, opened 2022-03-31
Script clean up

gurunrao:
Script clean up - remove unused run scripts and fix_classpath script.


scottmarlow (commented):
It was a good idea to add docker support but I agree we are not using this now and it is outdated.

scottmarlow (approved):

PR #919, opened 2022-04-05
issue 876, Update CDI bean usage for change in default bean-discovery-mode

scottmarlow:


gurunrao (approved):
LGTM

Issue #930, opened 2022-04-14
Expression Language test methodExpressionMatchingAmbiguousTest has wrong test input


arjantijms:
According to @markt-asf the purpose of this test is as follows:

```
In summary the target bean in the test below includes these methods:

public String targetE(Long, Long)
public String targetE(String, String)

By calling with a String and a numeric parameter the test is checking 
the following ambiguity:

Should the the String parameter be coerced to a number and targetE(Long, 
Long) called or should the numeric parameter be coerced to a String and 
targetE(String, String) called?
```

This aligns with EL:SPEC:80;

```
Identification of methods in method expressions from  parameter types follows the rules of the JLS with the addition
that coercion resolution is less specific than overload  resolution. [MethodExpressionMatch]
```

However the call with the test input is:

```java
bean.targetE('aaa',1234)
```

When it is said "Should the the String parameter be coerced to a number", it means that 'aaa' will be coerced to a number.

However,  'aaa' cannot be coerced to a number. Using the normal coercion code, e.g. `Long.valueOf(val)` with val "aaa", this would throw:

```java
java.lang.NumberFormatException: For input string: "aaa"
```

Therefore, the test seems wrong as it currently stands. I think it should be simply changed to:

```java
bean.targetE('1234',1234)
```

cc @pnicolucci

arjantijms:
Thanks @markt-asf

PR #931, opened 2022-04-14
Fix input of methodExpressionMatchingAmbiguousTest

arjantijms:
Fixes the test so that an Ambiguous situation occurs, which is what the test is testing the behaviour for.

**Fixes Issue**
https://github.com/eclipse-ee4j/jakartaee-tck/issues/930

**Describe the change**
For the coercion to Long (number) test, the string should represent a number. It was 'aaa' before, which is not a number, and has been changed into '1234', which does represent a number and can be coerced.

Signed-off-by: Arjan Tijms <arjan.tijms@gmail.com>


markt-asf (approved):
Makes sense to me and the test still passes for Tomcat.

Issue #932, opened 2022-04-14
New Expression Language tests don't setup ElContext for invoke


arjantijms:
Some of the new tests for the Jakarta Expression Language 5.0.0 TCK, like e.g.

`com/sun/ts/tests/el/api/jakarta_el/methodexpression/ELClient.java#methodExpressionMatchingCoercionBeatsExactVarArgsTest` contain code such as the following:

```
        StringBuffer buf = new StringBuffer();
        String exprStr = "#{bean.targetD('1','1')}";

        ExpressionFactory expFactory = ExpressionFactory.newInstance();
        ELContext context = (new VarMapperELContext(props)).getELContext();

        MethodsBean bean = new MethodsBean();
        ValueExpression ve = expFactory.createValueExpression(bean, MethodsBean.class);
        context.getVariableMapper().setVariable("bean", ve);

        MethodExpression me = expFactory.createMethodExpression(context, exprStr, String.class, null);
        me.invoke(...);
```

It seems this form of method expressions with provided parameters (the '1' and '1' here)) have never been tested before, and so the custom `ELContext` hasn't been set up to take this into account.

Expressly (the former GlassFish RI) use an ELResolver for the `me.invoke(...)` part. That's where it goes wrong, as the test sets up its own custom ELContext (VarMapperELContext) which provides custom ELResolvers that are not suitable for method invocation.

I think besides the custom `VariableELResolver()`, the standard `jakarta.el.BeanELResolver()` should be in the list of resolvers.

cc @markt-asf @alwin-joseph

scottmarlow:
Is this issue resolved now via https://github.com/eclipse-ee4j/jakartaee-tck/pull/933?

markt-asf:
Yes.

PR #933, opened 2022-04-14
Make sure the test ELContext has a resolver that can invoke methods.

arjantijms:
**Fixes Issue**
https://github.com/eclipse-ee4j/jakartaee-tck/issues/932

**Describe the change**
The expression language tests make use of a custom test `ELContext`, which only contains some minimal functionality for passing tests. Before expressions of the form `#{bean.targetD('1','1')}` have seemingly not been tested, and therefor this minimal `ELContext` did not take them into account.

This change adds the standard `jakarta.el.BeanELResolver` to the list of resolvers, so that if a test needs to invoke something, it will succeed for those implementations that make use of resolvers to invoke methods.

**Additional context**
I've not been able to build the TCK bundle itself and test it, but did test this exact change on a copied example.

Signed-off-by: Arjan Tijms <arjan.tijms@gmail.com>

cc @markt-asf  @pnicolucci

### DIFF
--- a/src/com/sun/ts/tests/el/common/elcontext/VarMapperELContext.java
+++ b/src/com/sun/ts/tests/el/common/elcontext/VarMapperELContext.java
@@ -54,6 +54,7 @@ public class VarMapperELContext extends ELContext {
   public ELResolver getELResolver() {
     ELResolver elResolver;
     this.compResolver.add(new VariableELResolver());
+    this.compResolver.add(new jakarta.el.BeanELResolver());
     elResolver = (ELResolver) compResolver;
 
     return elResolver;


markt-asf:
The fix looks good to me. I've built the TCK with this change and it still passes on Apache Tomcat 10.1.x.

I think we might want to take a wider look at the TCK after this release but that is a discussion for another time.

arjantijms:
> The fix looks good to me. I've built the TCK with this change and it still passes on Apache Tomcat 10.1.x.

Thanks! One extra question, what were the exact commands you used to build the (EL) TCK?

> I think we might want to take a wider look at the TCK after this release but that is a discussion for another time.

Indeed, there's a few things. I refactored that class itself for instance to this locally:

```
public class VarMapperELContext extends ELContext {

    private final VariableMapper varMapper;
    private final CompositeELResolver compResolver;

    public VarMapperELContext(Properties testProps) {
        compResolver = new CompositeELResolver();
        compResolver.add(new VariableELResolver());
        compResolver.add(new BeanELResolver());
        varMapper = getVariableMapperImpl(testProps);
    }

    @Override
    public ELResolver getELResolver() {
        return compResolver;
    }

    public ELContext getELContext() {
        return this;
    }

    @Override
    public VariableMapper getVariableMapper() {
        return varMapper;
    }

    @Override
    public FunctionMapper getFunctionMapper() {
        return null;
    }

    private Object getImplSpecificInstance(String classname) {
        try {
            return Class.forName(classname)
                        .getDeclaredConstructor()
                        .newInstance();
        } catch (Exception cnfe) {
            throw new IllegalStateException(cnfe);
        }
    }
    // and more
}
```

I don't know why it keeps adding resolvers every time `getELResolver` is called. It seems wrong to me, but don't want to make more changes at this time beyond the absolute necessary.

markt-asf:
> Thanks! One extra question, what were the exact commands you used to build the (EL) TCK?

```export WORKSPACE=/home/mark/repos/jakarta-tck
export ANT_HOME=/opt/apache-ant-1.9.12
export M2_HOME=/opt/apache-maven-3.6.3
export GF_BUNDLE_URL=https://download.eclipse.org/ee4j/glassfish/glassfish-7.0.0-SNAPSHOT-nightly.zip
export JDK11_HOME=/opt/java/temurin-jdk-11-latest

docker/build_standalone-tcks.sh el
```
> I don't know why it keeps adding resolvers every time `getELResolver` is called. It seems wrong to me, but don't want to make more changes at this time beyond the absolute necessary.

Yes, that raised an eyebrow here as well.

Issue #937, opened 2022-04-19
Remove use of Applet API


scottmarlow:
https://bugs.openjdk.java.net/browse/JDK-8256145 marked the Applet API as deprecated for Java SE 17.  This issue is for tracking the removal of use of Applets in the Platform TCK.  Some applet references in Platform TCK:

src/web/jsp/spec/core_syntax/actions/plugin/JspPluginAppletTest.jsp
src/com/sun/ts/tests/jsp/spec/core_syntax/actions/plugin/URLClient.java

scottmarlow:
Adding this for EE 10 in case we have time to handle this.  The advantage of removing the use of deprecated applets, is to prepare for when they are later removed in that fewer TCK failures will occur.

scottmarlow:
Note that this was addressed for `EE10` via https://github.com/eclipse-ee4j/jakartaee-platform/issues/298

scottmarlow:
Closed as the intent of removing the use of Applet API changed into [Deprecate jsp:plugin, jsp:params and jsp:fallback](https://github.com/eclipse-ee4j/jsp-api/pull/228) which was handled via https://github.com/eclipse-ee4j/jakartaee-tck/pull/940

PR #938, opened 2022-04-19
#937 Remove use of Applet API

scottmarlow:

**Fixes Issue**
https://github.com/eclipse-ee4j/jakartaee-tck/issues/937 in support of https://github.com/eclipse-ee4j/jakartaee-platform/issues/298

**Describe the change**
Applet is deprecated as of Java SE 17 and was removed from the EE 10 Platform SPEC.  This change is removing the tests that use the Applet APIs.

CC @markt-asf  @alwin-joseph @anajosep @arjantijms @cesarhernandezgt @dblevins @m0mus @edbratt @gurunrao @jansupol @jgallimore @kazumura @kwsutter @LanceAndersen @bhatpmk @RohitKumarJain @shighbar @gthoman @brideck @scottmarlow


markt-asf:
Given the removal of support for NPAPI by browsers, we probably need to look at deprecating and removing `<jsp:plugin>` and associated structures. I could get the work done for that fairly quickly (in the next 24-28 hours) but is it practical to get that done for Jakarta EE 10?

Given the specs will still support `<jsp:plugin>` where use of applet is valid, this change seems like the simplest way forward for the TCK.

starksm64:
I would say that is the correct thing to do. Remove it from the spec and update the spec documents in the PR.
https://github.com/jakartaee/specifications/pull/435

markt-asf:
Removal without prior deprecation would be unusual. I'd want to ensure there was wide support for that. It would also require moving to 4.0.0 which triggers a bunch of other processes. I'm not sure there is time to do this at this point.

Deprecation and a change to make the `<jsp:plugin>` a NO-OP could be done in the next 24 hours or so - including updating the TCK tests to ensure that they were NO-OPs (rather than removing the tests).

scottmarlow:
> Deprecation and a change to make the `<jsp:plugin>` a NO-OP could be done in the next 24 hours or so - including updating the TCK tests to ensure that they were NO-OPs (rather than removing the tests).

So the TCK tests would be updated to fail if the jsp:plugin is not ignored, that sounds good.

markt-asf:
PR for changes to Jakarta Pages: https://github.com/eclipse-ee4j/jsp-api/pull/228
Feedback welcome (on that PR please).

Issue #968, opened 2022-05-11
Task for addressing ejblitejsf vehicle failures


starksm64:
We need to follow the approach in the #962 PR for every set of tests that have the ejblitejsf vehicle mapping in the install/jakartaee/other/vehicle.properties file. This looks to be every test package under:

- [com/sun/ts/tests/jta/ee/transactional](https://github.com/eclipse-ee4j/jakartaee-tck/issues/977)
- com/sun/ts/tests/ejb30/lite except
  - com/sun/ts/tests/ejb30/lite/packaging/embed
  - com/sun/ts/tests/ejb30/lite/stateful/timeout
  - com/sun/ts/tests/ejb30/bb/session/stateful/timeout
- com/sun/ts/tests/ejb30/bb/async (Covered in https://github.com/eclipse-ee4j/jakartaee-tck/pull/1005 )
- com/sun/ts/tests/ejb30/bb/session/stateful/concurrency
- [com/sun/ts/tests/ejb32/lite](https://github.com/eclipse-ee4j/jakartaee-tck/issues/968) (done via https://github.com/eclipse-ee4j/jakartaee-tck/pull/971 and https://github.com/eclipse-ee4j/jakartaee-tck/pull/1008 )

I'll create sub issues for each package so we can track who is doing the work.

alwin-joseph:
Status of ejb test failures related to migration of  ejblitejsf vehicle : 

CI job : https://ci.eclipse.org/jakartaee-tck/job/10/job/eftl-jakartaeetck-run-100/68/ 
Pipeline job : https://ci.eclipse.org/jakartaee-tck/blue/organizations/jenkins/jakartaee-tck/detail/master/1824/pipeline 


1. ejb30/lite/async : 

From failures logs   - 

```
   [runcts] OUT => [javatest.batch] 05-19-2022 09:50:49:  Harness - TSScript:  No vehicle ear found in : "/root/jakartaeetck/bin/xml/../../dist/com/sun/ts/tests/ejb30/lite/async/singleton/annotated".  If this is a test that requires test clients to be built before executing the tests, please refer to the TCK documentation for further instructions.
   [runcts] OUT => [javatest.batch] Failed. TSScript: No vehicle ear found in : "/root/jakartaeetck/bin/xml/../../dist/com/sun/ts/tests/ejb30/lite/async/singleton/annotated". If this is a test that requires test clients to be built before executing the tests, please refer to the TCK documentation for further instructions.
```

[...]

PR #962, opened 2022-05-06
Convert jsp to xhtml for ejblitejsf vehicle test runs - ejb30/lite/appexception

alwin-joseph:
**Describe the change**
Initial change to fix the ejb30/lite/appexception tests in ejblitejsf vehicle by converting .jsp to .xhtml. 

The change does fixes the failures in ejblitejsf vehicle but causes failures in ejbliteservlet & ejbliteservlet2 vehicles. 

CC @scottmarlow

PS : The changes in Client files with added annotation of `@jakarta.inject.Named` & `@jakarta.enterprise.context.RequestScoped` is causing test failures in ejbliteservlet & ejbliteservlet2 vehicles. 
The ejbliteservlet & ejbliteservlet2 vehicles uses jsp too (see https://github.com/eclipse-ee4j/jakartaee-tck/tree/master/src/com/sun/ts/tests/common/vehicle/ejbliteservlet2). 

@arjantijms Is there a way to use the same Client (eg. [src/com/sun/ts/tests/ejb30/lite/appexception/stateful/annotated/Client.java](https://github.com/eclipse-ee4j/jakartaee-tck/blob/master/src/com/sun/ts/tests/ejb30/lite/appexception/stateful/annotated/Client.java) ) with annotations `@jakarta.inject.Named` & `@jakarta.enterprise.context.RequestScoped` for ejbliteservlet & ejbliteservlet2 vehicles that has jsp.

Is using a separate client (say JsfClient) for  ejblitejsf vehicle better? 

Update : The latest changes addresses all failures in ejb30/lite/appexception using the recommendations provided.
There are more changes required to fix the same failures in other suites by having separate Clients for ejblitejsf vehicle tests. Those will be addressed in subsequent PRs.

PR #971, opened 2022-05-11
Address ejblitejsf test failures in ejb32 - part1

alwin-joseph:
**Related Issue(s)**
https://github.com/eclipse-ee4j/jakartaee-tck/issues/968

**Describe the change**
Migrates .jsp to .xhtml, Managed Beans to Named Beans , new Client for ejblitejsf vehicle tests (configuration change in install/jakartaee/other/vehicle.properties) in ejb32 suite 

**Additional context**
- This PR extends the  changes in https://github.com/eclipse-ee4j/jakartaee-tck/pull/962 to ejb32 suite.
- Please note this PR reverts a change in 962 (file bin/xml/ts.vehicle.xml), to exclude the client classes from archives . eg: https://github.com/eclipse-ee4j/jakartaee-tck/blob/54b9d6987e47f265391152961740742d0ca7e0d2/bin/xml/ts.vehicles.xml#L475 is changed in this PR. 
- This PR covers all ejblitejsf failures in ejb32/ suite except below. These will be covered in a separate PR.
src/com/sun/ts/tests/ejb32/lite/timer/schedule/auto/attr/singleton
src/com/sun/ts/tests/ejb32/lite/timer/schedule/auto/attr/stateless

CC @starksm64 @arjantijms @scottmarlow @gurunrao


starksm64:
So both the Client.class and JsfClient.class should be included in the vehicle archive?

arjantijms:
> So both the Client.class and JsfClient.class should be included in the vehicle archive?

That should be the case. Client is what the Servlets inherit from. JsfClient uses a scope and is therefor not compatible with anything a Servlet can inherit from.

I would suggest to use the name "FacesClient", as we are phasing out the term "JSF'.

scottmarlow (approved):

PR #1062, opened 2022-06-09
exclude CDI_SE tests when running JSON-B TCK

scottmarlow:

Workaround GlassFish 7 Web Profile not support CDI_SE tests

https://ci.eclipse.org/jakartaee-tck/job/10/job/jakarta-jsonb-tck-glassfish-run/lastCompletedBuild/testReport shows zero failures running JSON-B TCK with GlassFish 7 nightly build web profile.

**Fixes Issue**

https://github.com/eclipse-ee4j/jsonb-api/issues/326

### DIFF
--- a/glassfish-runner/jsonb-tck/pom.xml
+++ b/glassfish-runner/jsonb-tck/pom.xml
@@ -89,6 +89,9 @@
               <additionalClasspathElement>${project.build.directory}/${glassfish.toplevel.dir}/glassfish/modules/parsson.jar</additionalClasspathElement>
               <additionalClasspathElement>${project.build.directory}/${glassfish.toplevel.dir}/glassfish/lib/appclient/weld-se-shaded.jar</additionalClasspathElement>
           </additionalClasspathElements>
+          <excludes>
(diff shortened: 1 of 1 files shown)


scottmarlow:
I mentioned this change on https://github.com/eclipse-ee4j/jsonb-api/issues/326, lets see if the JSON-B team accepts this change as a valid workaround before merging.

m0mus:
JSONB TCK has two budles: the standalone bundle using JUnit and the platform bundle using JavaTest. The standalone bundle is designed for TCK testing a JSONB implementation, the platform bundle is designed to test a platform implementation. The reuirement is that JSONB implementation used in the platform is certified (passed the JUnit bundle).

What you are trying to do here is to extract JSONB implementation out of the platform and run JUnit bundle on it using the platform class/module path. It's not the way how JSONB TCKs are designed.

If Yasson is used (which is JSONB compliant) you don't need to run JUnit bundle at all. Run only the platform bundle. If other implementation is used, it must be JSONB certified first. It means that implementation authors should run JUnit bundle on it. It should not use any all server class/module path though. It also should provide all dependencies required by JUnit bundle including CDI SE.

PR #1076, opened 2022-06-16
Use MQ distribution zip directly

pzygielo:
**Fixes Issue**

**Related Issue(s)**

**Describe the change**
I propose for consideration for standalone Messaging TCK to use OpenMQ archive directly instead of GlassFish one.

**Additional context**
My understanding is that there is no need to bring whole GF for this TCK.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -169,6 +169,9 @@ spec:
 	  string(name: 'OLD_GF_BUNDLE_URL', 
            defaultValue: '', 
            description: 'URL required for downloading Old GlassFish Full/Web profile bundle' )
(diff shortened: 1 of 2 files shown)


scottmarlow:
Soon we will be finished with EE 10 and able to start moving to separate technology TCKs to be build/run via (technology specific) Maven pom within the Platform TCK.  Perhaps Messaging could start doing that.  I think the idea is to have a maven pom that supports different profiles (e.g. Full Platform, Web Profile (if needed), Java SE).  If you add an Arquillian TCK Runner for MQ, that helps ensure MQ can just use that.  Hope this helps.

PR #1093, opened 2022-07-06
Use Exousia JACC package

Pandrex247:
Signed-off-by: Andrew Pielage <andrew.pielage@payara.fish>

**Fixes Issue**
None raised.

**Related Issue(s)**
None

**Describe the change**
The package name is wrong and the `RunCTS#enableJacc()` ant task is hardcoded to use the default GlassFish config file. 
This means when a runner uses this predefined task to execute the JACC TCK the server will fail with ClassNotFoundExceptions due to expecting the old `com.sun` class.

**Additional context**
None

### DIFF
--- a/bin/xml/impl/glassfish/common.xml
+++ b/bin/xml/impl/glassfish/common.xml
@@ -1421,7 +1421,7 @@
          </antcall>
 
          <antcall target="create-jvm-options" >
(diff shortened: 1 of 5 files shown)


arjantijms (approved):
That is the correct provider. It's the one normally defined in domain.xml in GF.

PR #1129, opened 2022-10-24
copy Alwin Joseph new wiki on Standard Rules while separating standalone TCKs from platform TCK to (new) notes folder that any user can create pull requests from.  The idea being the well known wiki could switch to using notes added to the repo over time.

scottmarlow:

I think it would be more natural to store design notes directly in the Platform TCK repo, so trying that via this pr.


alwin-joseph (approved):
+1

PR #1164, opened 2023-04-17
EL TCK Refactoring using Junit

alwin-joseph:
**Related Issue(s)**
https://github.com/jakartaee/platform-tck/issues/1126

**Describe the change**
- Converted all standalone legacy EL Tests to Junit (No arquillian used). 
- Used slf4j logging instead of TestUtil class
- Moved the tests from  https://github.com/eclipse-ee4j/jakartaee-tck/tree/tckrefactor/src/com/sun/ts/tests/el to el/ folder 
- Created glassfish runner for the same el tck

Pending: 
- Use servlet & JSP containers to run these tests for platform [vehicle.properties](https://github.com/jakartaee/platform-tck/blob/e3ab547f4308877e2178ef21609c0d8aac45f766/install/jakartaee/other/vehicle.properties#L80) 
- Refactor standalone EL signature tests  
- Copy the test in https://github.com/jakartaee/platform-tck/pull/1159 also.


alwin-joseph:
I see PR https://github.com/jakartaee/platform-tck/pull/1159 on master branch that adds a new EL test. Reminder to self that the same should be merged to tckrefactor too along with this PR or as a separate change.

el/pom.xml
```diff
@@ -46,6 +46,16 @@
             <groupId>jakarta.el</groupId>
             <artifactId>jakarta.el-api</artifactId>
         </dependency>
+       <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter</artifactId>
+            <scope>compile</scope>
+        </dependency>
+        <dependency>
```
dmatej, line 54: Would it be possible to avoid dependency on SLF4J and use System.Logger instead?
See https://www.slf4j.org/manual.html#jep264

alwin-joseph: Done.

scottmarlow:
> Hi @starksm64
> 
> With respect to the vehicle implementation, would you have any comments for EL TCK refactoring in this PR.
> 
>     1. Basically, the EL TCK in standalone mode is refactored in this PR and uses only JUnit. For legacy EL standalone TCK the tests did not run in any vehicle (https://github.com/jakartaee/platform-tck/blob/e3ab547f4308877e2178ef21609c0d8aac45f766/install/el/other/vehicle.properties#L52
>        ) and hence the changes in this PR does seem sufficient to me.
> 
>     2. For platform TCK, the EL tests run in 2 vehicles (https://github.com/jakartaee/platform-tck/blob/e3ab547f4308877e2178ef21609c0d8aac45f766/install/jakartaee/other/vehicle.properties#L80
>        ) servlet & jsp.
>        Com.sun.ts.tests.common.vehicle.jsp/servlet has the legacy implementation of jsp/servlet vehicles (same folder with other vehicles). I believe the requirement will be to duplicate the current vehicle implementation or equivalent scenario using Arquillian.

./webartifacts/servlet/src/main/resources/api/jakarta_servlet/servletcontext40/addJspFile.jsp is currently referenced from the refactored Servlet TCK but I don't see any references from the Servlet TCK to `Servlet.*Vehicle` or `JSP.*Vehicle`.  So we might want to ask @olamy for input on whether we should do the same for other `Web only` container using Platform TCK tests.  For the Standalone EL tests, we shouldn't reference the Web container.

> 
>     3. I found that batch TCK created ejb vehicle and runner in the PR [Add EJB runner to platform/Arquillian path as a separate failsafe execution eclipse-ee4j/batch-tck#40](https://github.com/eclipse-ee4j/batch-tck/pull/40).

This is a good question as we currently have the problem of ratifying multiple Specifications before `Core`/`Web Profile` + `Full Platform` Specifications are complete.

>        Should we also follow the same approach? 
IMO, it is better to not follow the same approach.  We really should have separate source/maven artifacts for the Standalone (Java SE only tests) and the profile level TCK tests.  I think that  the profile level TCK tests can and should leverage the Standalone tests to avoid duplication (Profile TCK developers should update/adjust their tests accordingly to account for changes in the underlying Specification TCK tests that may occur prior to said underlying Specification being ratified).

> Any idea if arquillian already has extensions or possibilities to extend the test runs in custom vehicles or containers.

We have a few choices on how we deal with each of the vehicles.  For reference the different vehicles are still in https://github.com/jakartaee/platform-tck/tree/tckrefactor/common/src/main/java/com/sun/ts/tests/common/vehicle.  I think we should walk through how each are used and consider if there is value in using them.  

I haven't gotten to the Persistence tests yet but I will be looking at `5.4.14.3 Persistence Test Vehicles` in https://github.com/jakartaee/platform-tck/blob/tckrefactor/user_guides/src/main/jbake/content/config.adoc (stateless3, stateful3, appmanaged, appmanagedNoTx, pmservlet, puservlet) vehicles.  

> 
> 
> cc @scottmarlow @gurunrao

el/pom.xml
olamy: Why?
Better to have something without real implementations and user can choose whatever implementations they prefer.
If you prefer `System.Logger` fine you add this slf4j dependency at runtime, if another user prefere log4j or logback fine they just add the correct slf4j binding at runtime.
Here you force user with an implementation they may not want.
slf4j have the advantage of being implementation neutral and users have the freedom to choose the one they prefer.

dmatej: SLF4J is NOT neutral, it is an additional dependency which is not necessary these days. System.Logger is the neutral API, which is inside Java since Java 9. SLF4J2 can bind to it too just like other logging frameworks.
See also https://openjdk.org/jeps/264 - the SLF4J is even explicitly mentioned.

olamy: oh right thanks!
 I didn't look at the other way, another cascade :) `System.Logger` -> `slf4j api` -> `final impl` 
that's a bit of layers but that's fine.

el/src/main/java/com/sun/ts/tests/el/api/jakarta_el/arrayelresolver/ELClientIT.java
```diff
@@ -33,26 +32,48 @@
...
   private Properties testProps;
 
   private static final String[] names = { "doug", "nick", "roger", "ryan",
       "ed" };
 
-  public static void main(String[] args) {
-    ELClient theTests = new ELClient();
-    Status s = theTests.run(args, System.out, System.err);
-    s.exit();
+  // public static void main(String[] args) {
```
gurunrao, line 52: good to remove jtest related unused code.

el/src/main/java/com/sun/ts/tests/el/api/jakarta_el/compositeelresolver/ELClientIT.java
```diff
@@ -37,26 +37,46 @@
...
 
-public class ELClient extends ServiceEETest {
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInfo;
+
+import java.lang.System.Logger;
+
+public class ELClientIT extends ServiceEETest {
```
gurunrao, line 47: ServiceEETest has code related to harness, do we require for Arquillian + Junit runs?

el/src/main/java/com/sun/ts/tests/el/api/jakarta_el/elcontext/ELClientIT.java
```diff
@@ -34,27 +34,42 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInfo;
+
+import java.lang.System.Logger;
+
+public class ELClientIT extends ServiceEETest {
+
+  private static final Logger logger = System.getLogger(ELClientIT.class.getName());
 
   private Properties testProps;
```
gurunrao, line 48: testProps can be removed if unused.

el/src/main/java/com/sun/ts/tests/el/api/jakarta_el/arrayelresolver/ELClientIT.java
alwin-joseph: Done

el/src/main/java/com/sun/ts/tests/el/api/jakarta_el/elcontext/ELClientIT.java
alwin-joseph: Done

gurunrao (approved):

olamy (approved):

olamy (approved):

markt-asf:
This looks great. I managed to build the EL TCK locally without too much difficulty. `javatest.jar` was the only issue and that just needed to be installed at the right coordinates in my local repository.

I was able to configure the TCK to run against Tomcat's EL implementation easily. It was also simple to swap between the Jakarta EL API JAR and the EL API JAR provided by Tomcat. Everything passed in all combinations.

The tests report a total of 359 tests which I assume is one less than the 360 reported by the current TCK as the signatures are not tested. Is that correct?

Finally, by far the most impressive thing about this change is the time taken to run the TCK. The current TCK takes tens of minutes to complete. This refactored version takes less than 5s.

scottmarlow:
@markt-asf I think the signature test is still pending.

alwin-joseph:
> 
> The tests report a total of 359 tests which I assume is one less than the 360 reported by the current TCK as the signatures are not tested. Is that correct?
> 
Correct. The signature test work is pending (The total tests for standalone EL TCK are 349 +1 signature test).

PR #1178, opened 2023-05-26
ClassLoader issue in embedded Arquillian

jbescos:
**Fixes Issue**
https://github.com/helidon-io/helidon/issues/6884

**Related Issue(s)**
https://github.com/helidon-io/helidon/issues/6884

**Describe the change**
Using embedded Arquillian, there is only one JVM, so next line will not get the resource from the web application, it will get the test execution classpath (the file will not be found):
`getClass().getResourceAsStream("/key.pub")
`
This suggested fix is more flexible, because the web container is able to set its own class loader.

**Additional context**
I faced this issue integrating this in Helidon.

@starksm64 @scottmarlow

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ target/
 .project
 /glassfish-runner/batch-tck/apitests/test.properties
 .classpath
+.factorypath
 .settings/
 classes/
 dist/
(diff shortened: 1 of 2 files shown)


dmatej (approved):

arjantijms:
p.s. for Piranha Cloud and Arquillian I worked on an isolated class loader so tests won't pick up thing from the environment. Might be worthwhile to see if that can be applied to this situation as well at some point.

PR #1188, opened 2023-07-31
 ClassLoader issue in embedded Arquillian (additional null check)

jbescos:
**Fixes Issue**
https://github.com/jakartaee/platform-tck/issues/1187#issuecomment-1655868687

**Describe the change**
In case `ClassLoader classLoader = Thread.currentThread().getContextClassLoader();` returns null, it will use the previous approach.

**Additional context**
This is because of one comment of @scottmarlow 

@starksm64 @scottmarlow


dmatej (approved):
I would probably extract that to a private method and inverted the condition order, but it is ok this way.

scottmarlow (approved):

scottmarlow:
@jbescos thanks!

PR #1195, opened 2023-10-13
[763] Add missing exclusions from the CDI TCK's tck-tests.xml to the …

### LINKED ISSUE #1196: [tck-challege] Jakarta EE Core TCK's cdi-lite-tck-suite.xml out of sync with CDI's tck-tests.xml
**Relevant Information**

1. The relevant specification version and section number(s).
    - Jakarta EE 10 Core Profile
2. The coordinates of the challenged test(s).
    - `jakarta.ee.tck.coreprofile:cdi-lite-tck-suite:10.0.1` 
3. The exact TCK version.
    - 10.0.1
4. The implementation being tested, including name and company.
    - WildFly by Red Hat
5. A full description of why the test is invalid and what the correct behavior is believed to be.
    - There are 3 CDI tests that should be excluded from the Jakarta EE Core TCK's CDI Lite TCK that are not:
        - `org.jboss.cdi.tck.tests.implementation.simple.lifecycle.SimpleBeanLifecycleTest`
        - `org.jboss.cdi.tck.tests.context.DestroyForSameCreationalContextTest`
        - `org.jboss.cdi.tck.tests.definition.bean.types.ManagedBeanTypesTest`
6. Any supporting material; debug logs, test output (.jtr file), test logs, run scripts, etc.

The [`cdi-lite-tck-suite.xml`](https://github.com/jakartaee/platform-tck/blob/82a976ee03a7ea47bcd8ff8623579708802e937f/core-profile-tck/cdi-tck-suite/src/main/resources/cdi-lite-tck-suite.xml) is out of sync with CDI's [`tck-tests.xml`](https://github.com/jakartaee/cdi-tck/blob/0c25a2d9cd8206cf7ff30184884e6ce062fb3176/impl/src/main/resources/tck-tests.xml#L18). These tests should be excluded in the `cdi-lite-tck-suite.xml`:
```xml
<!-- https://github.com/jakartaee/cdi-tck/issues/453 -->
<class name="org.jboss.cdi.tck.tests.implementation.simple.lifecycle.SimpleBeanLifecycleTest">
    <methods>
        <exclude name="testCreateReturnsSameBeanPushed"/>
    </methods>
</class>
<class name="org.jboss.cdi.tck.tests.context.DestroyForSameCreationalContextTest">
    <methods>
        <exclude name="testDestroyForSameCreationalContextOnly"/>
    </methods>
</class>

<!-- https://github.com/jakartaee/cdi-tck/issues/485 -->
<class name="org.jboss.cdi.tck.tests.definition.bean.types.ManagedBeanTypesTest">
    <methods>
        <exclude name=".*"/>
    </methods>
</class>
```

The latter (`org.jboss.cdi.tck.tests.definition.bean.types.ManagedBeanTypesTest`) causes issues with the TCK passing on Java 21.

If we cannot get this merged and a new Jakarta EE 10 Core Profile TCK released, we're asking that we can explicitly exclude these tests in our test configuration. In our case it will be on the `maven-surefire-plugin` section of our `pom.xml` for the CDI Lite TCK tests.

**Additional context**
The CDI TCK Challenge issues:
- https://github.com/jakartaee/cdi-tck/issues/483
- https://github.com/jakartaee/cdi-tck/issues/485

jamezp:
…Jakarta EE Core Profiles cdi-lite-tck-suite.xml.

**Fixes Issue**
Resolves #1196 

**Describe the change**
These exclusions are in the CDI TCK, however they were not migrated to the CDI Lite TCK exclusions.

**Additional context**
This fix will allow the Jakarta EE Core Profile to pass, assuming the container passes :), with Java 21.

@starksm64 @scottmarlow

CC @manovotn @Ladicek

### DIFF
--- a/core-profile-tck/cdi-tck-suite/src/main/resources/cdi-lite-tck-suite.xml
+++ b/core-profile-tck/cdi-tck-suite/src/main/resources/cdi-lite-tck-suite.xml
@@ -34,5 +34,25 @@
             <package name="org.jboss.cdi.tck.tests.*" />
             <package name="org.jboss.cdi.tck.interceptors.tests.*" />
         </packages>
+        <classes>
+            <!-- https://github.com/jakartaee/cdi-tck/issues/453 -->
+            <class name="org.jboss.cdi.tck.tests.implementation.simple.lifecycle.SimpleBeanLifecycleTest">
+                <methods>
+                    <exclude name="testCreateReturnsSameBeanPushed"/>
+                </methods>
+            </class>
+            <class name="org.jboss.cdi.tck.tests.context.DestroyForSameCreationalContextTest">
+                <methods>
+                    <exclude name="testDestroyForSameCreationalContextOnly"/>
+                </methods>
+            </class>
+
+            <!-- https://github.com/jakartaee/cdi-tck/issues/485 -->
+            <class name="org.jboss.cdi.tck.tests.definition.bean.types.ManagedBeanTypesTest">
+                <methods>
+                    <exclude name=".*"/>
+                </methods>
+            </class>
+        </classes>
     </test>
 </suite>


jamezp:
I've noticed I filed the issue under the wrong project, platform instead of platform-tck. I'll remedy that now and update this PR.

scottmarlow (approved):

scottmarlow:
@manovotn @Ladicek what do you think of the change?

CC @starksm64

manovotn (approved):
> @manovotn @Ladicek what do you think of the change?
> 
> CC @starksm64

Frankly, I think it's unfortunate that this file is in a completely different repository as it will be hard to keep in sync - the TCK challenges can modify the original file fairly easily so this situation is likely to repeat.
Would it be possible to instead use the original file? Genuine question as I know that it can (and currently does) contain exclusions for tests that aren't part of the core, but I am not sure if that would be an issue for testng?

As for the content of the PR, that looks good.

jamezp:
I think the one issue with using the original file is are the test group definitions. The `cdi-lite-tck-suite.xml` has:
```xml
<groups>
    <run>
        <exclude name="cdi-full"/>
        <exclude name="se"/>
    </run>
</groups>
```

I'm not sure how you can add exclude groups like this.

Another option, for the TCK itself, would be to use the `@Test(enabled = false)`. Not as ideal I know, but it would be more "global".

starksm64:
Historically the exclusions list could be shipped without an update to the TCK binary, so excluding at the test level was not always viable. The current TCK rules require a full respin of the TCK binary, so maybe excluding at the source level would be better.

In terms of exclusion groups, that is the syntax described in the TestNG docs:
https://testng.org/doc/documentation-main.html#exclusions

manovotn:
> I think the one issue with using the original file is are the test group definitions. The `cdi-lite-tck-suite.xml` has:

Ah true, didn't realize that.

> Historically the exclusions list could be shipped without an update to the TCK binary, so excluding at the test level was not always viable. The current TCK rules require a full respin of the TCK binary, so maybe excluding at the source level would be better.

The XML has the notable upside of being one place to track all exclusions. TCK actually had two source exclusions for last few years that nobody really knew about (https://github.com/jakartaee/cdi-tck/pull/482).

Maybe it would be enough to create some how-to-release doc for TCK that also mentions checking the exclusion file here?

PR #1209, opened 2023-12-12
TCK Refactor of JSP & EL Standalone Signature tests 

alwin-joseph:
**Related Issue(s)**
https://github.com/jakartaee/platform-tck/issues/1126

**Describe the change**
This converts the signature tests in JSP & EL standalone TCKs to JUnit tests.
Below tests are moved to the respective standalone TCK folders:
- signaturetest/src/main/java/com/sun/ts/tests/signaturetest/el/ELSigTest.java
- signaturetest/src/main/java/com/sun/ts/tests/signaturetest/jsp/JSPSigTest.java

The TCK runners are modified to pick up these tests for execution.

There was mention about pending conversion of signature tests in below related PRs:
- https://github.com/jakartaee/platform-tck/pull/1164 
- https://github.com/jakartaee/platform-tck/pull/1182


gurunrao:
IMO, having a maven runner for standalone Signature tests should be sufficient. Maven runner will be easy for vendors to understand.
Also since signature tools have ability to ignore the JDK classes, we should be able to remove dependency on jimage tool and respective class dump for signature test run.

scottmarlow (approved):

alwin-joseph:
> IMO, having a maven runner for standalone signature tests should be sufficient. Maven runner will be easy for vendors to understand. Also since signature tools have ability to ignore the JDK classes, we should be able to remove dependency on jimage tool and respective class dump for signature test run.

Apparently the maven runner of the sigtest-maven-plugin uses target "check" from https://github.com/eclipse-ee4j/jakartaee-tck-tools/blob/master/tools/sigtest/src/main/java/org/netbeans/apitest/SigtestCheck.java only.

The legacy signature test runner uses https://github.com/jtulach/netbeans-apitest/blob/master/src/main/java/com/sun/tdk/signaturetest/SignatureTest.java which has static and reflective modes of running the signature test.

IMO we can continue to use the legacy way for now until further improvement is made to the sigtest tool to execute  src/main/java/com/sun/tdk/signaturetest/SignatureTest via maven. So currently we do use the sigtest-maven-plugin , but the legacy framework is made to call the plugin with appropriate parameters.

PR #1221, opened 2024-02-07
switch GroupId to jakarta.tck instead of jakartatck

scottmarlow:
As per discussion on mailing list and Platform TCK call today (February 7, 2024) we should use `jakarta.tck` as the groupId for Platform TCKs.

### DIFF
--- a/appclient/pom.xml
+++ b/appclient/pom.xml
@@ -21,7 +21,7 @@
     <modelVersion>4.0.0</modelVersion>
 
     <parent>
-        <groupId>jakartatck</groupId>
+        <groupId>jakarta.tck</groupId>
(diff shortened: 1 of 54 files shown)


scottmarlow:
For reference the change was made locally via:

> find -name pom.xml -type f| xargs sed -i 's/jakartatck/jakarta.tck/g'

olamy (approved):
LGTM

alwin-joseph (approved):
+1

alwin-joseph:
The new annotations module and glassfish-runner/annotations-tck also has pom.xml that needs change too. These were merged after this PR.

dmatej (approved):

PR #1232, opened 2024-03-01
Persistence tck bundle, package rename, documentation, signature tests.

gurunrao:
Persistence tck bundle, package rename, documentation, signature tests.


jpa/pom.xml
```diff
@@ -23,25 +23,45 @@
...
-    <name>JPA</name>
-    <description>JPA</description>
+    <artifactId>persistence-tck</artifactId>
+    <version>3.2.0</version>
+    <packaging>pom</packaging>
+    <name>persistence</name>
+    <description>persistence tck tests</description>
+    <modules>
+        <module>common</module>
+        <module>dist</module>
```
lukasj, line 35: `dist` looks missing in the PR

gurunrao: thanks @lukasj, `git status` was not showing the folder under untracked files, I have added folder using `git add -f`

lukasj: I recommend renaming the folder to `distribution` (or whatever you find suitable) instead - `dist` is often being put to ignore list of VCS (because of `ant`) and one can easily get surprised by local changes not being part of the PR (they are easy to miss)

PR #1236, opened 2024-03-02
set version to 11.0.0-SNAPSHOT, set src level to 17

lukasj:
set project version to 11.0.0-SNAPSHOT and `maven.compiler.release` to 17 (as EE 11 is supposed to have use that as its minimum)

@starksm64 @scottmarlow


.github/workflows/maven.yml
```diff
@@ -32,19 +32,19 @@ jobs:
...
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       - name: Set up JDK
-        uses: actions/setup-java@v3
+        uses: actions/setup-java@v4
         with:
-          java-version: 11
+          java-version: 17
```
pzygielo, line 47: Perhaps `matrix.java` could be used instead.

websocket/pom.xml
```diff
@@ -84,12 +82,12 @@
             <build>
                 <plugins>
                     <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
```
pzygielo, line 85: Isn't the implicit groupId good enough?

.github/workflows/maven.yml
lukasj: then just do it ;-)

websocket/pom.xml
lukasj: I prefer explicit usage of gId

pzygielo: you are already changing the area, improve for the sake of DRY

.github/workflows/maven.yml
pzygielo: you are already changing the area, improve for the sake of DRY

lukasj: GH has Suggest a change feature which I can just accept.

lukasj:
so at this point I believe the GH job will pass with either `-Pstaging` or Servlet 6.1.0-M1 TCK being available in central

Notable changes in this PR to be able to get to this state:
* application of spotless reordered executions for running Batch TCK - I resurrected original ordering by adding _Step 0X_ prefix to execution IDs to make the TCK as well as spotless check happy
* jsonp TCK outputs `\uFFFF` character which is being captured to report xml files during test run, but the report plugin fails to handle this character, therefore I disabled the report transformation part (AFAICT usage of `\uFFFF` is not guaranteed to be interoperable)
* `jsp/pom.xml` had version set to `4.0.0`, I changed that to `11.0.0-SNAPSHOT` as that is what I believe it should be as long as it is included in the main project; Should the version of this project remained untouched, then the resolution of this is to remove it from the main pom, so it is not directly included there

lukasj:
@scottmarlow there's a green tick now :-) It should be there also after the most recent commit, so I believe this is ready for merge (even though it is far from being perfect, but even in this state it will allow me to continue on what I wanted to work on Fri - TCK UG for persistence...)

scottmarlow (approved):
Thanks Lukas!  The [matrix](https://github.com/actions/setup-java) change looks great as well!

PR #1242, opened 2024-03-05
let GH run just the last commit and cancel build in progress

lukasj:
@scottmarlow consider adding this so there's only the last commit being checked by GH action build

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -26,6 +26,10 @@ permissions:
   id-token: write
   checks: write
 
+concurrency:
+  group: ${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   build:
 


.github/workflows/maven.yml
```diff
@@ -26,6 +26,10 @@ permissions:
   id-token: write
   checks: write
 
+concurrency:
+  group: ${{ github.ref }}
+  cancel-in-progress: true
```
scottmarlow, line 31: nice, looking at https://docs.github.com/en/actions/using-jobs/using-concurrency now...

So this applies to lukasj:gh_cancel branch in this case such that if you push another commit, any previous still running test is stopped so that a new test is started.  Very good idea!

scottmarlow (approved):

PR #1258, opened 2024-03-18
sync persistence challenges/fixes from master branch

gurunrao:
sync persistence challenges/fixes from master branch

### DIFF
--- a/jpa/docs/TCK-Exclude-List.txt
+++ b/jpa/docs/TCK-Exclude-List.txt
@@ -18,6 +18,26 @@
 # This is intended only for documentation purpose and is not used to exclude any tests.
 # The tests are excluded from the test source using @Disabled tag in JUnit framework.
 #
-ee/jakarta/tck/persistence/se/schemaGeneration/annotations/tableGenerator/Client.java#tableGeneratorTest
-ee/jakarta/tck/persistence/core/query/language/Client6.java#resultContainsFetchReference
+
+#
 ee/jakarta/tck/persistence/core/metamodelapi/identifiabletype/Client.java#getDeclaredSingularAttributes
+
+#
+#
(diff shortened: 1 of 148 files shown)


gurunrao:
@beikov - Please review.

beikov:
Sorry, but I have no idea what the current state of the Persistence TCK is, so I don't know whether these excludes are ok or not.

lukasj (approved):
consider updating the text in the TCK-Exclude-List.txt file, currently it says (on line 17) _This file lists the excluded tests from Jakarta EL TCK._ - that does not look correct

gurunrao:
> consider updating the text in the TCK-Exclude-List.txt file, currently it says (on line 17) _This file lists the excluded tests from Jakarta EL TCK._ - that does not look correct

Thanks, was copy and base error.

scottmarlow:
> Sorry, but I have no idea what the current state of the Persistence TCK is, so I don't know whether these excludes are ok or not.

Some guidance on excludes from https://jakarta.ee/committees/specification/tckprocess:
> Excludes
>
> Excludes MUST be included in the TCK project release in a format that is compatible with the testing framework in use so that as the excludes are updated, the affected tests are automatically removed from the test suite.
> 
> Adding excluded tests
> Excluded tests should be added back in for every major Jakarta EE release by emptying the test exclude list for every Specification developed in the respective major EE release.

Should we empty all of the TCK excludes list for minor specification updates in EE 11?  Or should that only be done for a major Persistence specification release?

I think that the underlying message behind emptying the exclude list is to give the relevant specification team a chance to rewrite the excluded tests to make them portable or solve whatever problem caused them to be excluded.  I don't think we have process setup currently to deal with this effectively.

PR #1268, opened 2024-03-26
prepare for 11.0.0-M1 release of signaturetest, runtime, libutil, common

scottmarlow:


gurunrao (approved):
LGTM

gurunrao:
Build is failing due to dependency on signaturetest, runtime, libutil, common libraries SNAPSHOT version in other modules like assembly. IMO for build to succeed, we should search and replace all dependencies to M1.

olamy:
usually in the Maven world poms should have non SNAPSHOT versions only temporarily. just the time to commit, tag then commit to next SNAPSHOT.
I don't think it's a good idea to commit such change in the main branch. Users who get this revision locally may have wrong version 11.0.0-M1 locally.

Should be in a release branch, not the main working one.

scottmarlow:
> Build is failing due to dependency on signaturetest, runtime, libutil, common libraries SNAPSHOT version in other modules like assembly. IMO for build to succeed, we should search and replace all dependencies to M1.

+1 thanks @gurunrao 

> usually in the Maven world poms should have non SNAPSHOT versions only temporarily. just the time to commit, tag then commit to next SNAPSHOT.
I don't think it's a good idea to commit such change in the main branch. Users who get this revision locally may have wrong version 11.0.0-M1 locally.
>
> Should be in a release branch, not the main working one.

@olamy should we create a `11.0.0-M1` branch for the release?

scottmarlow:
I pushed the release branch to https://github.com/jakartaee/platform-tck/tree/11.0.0-M1

olamy:
sorry for late reply but yes using a release branch is perfect

PR #1269, opened 2024-03-26
Update EJB Vehicles to EJB 3.x

brideck:
This takes care of a majority of the failures that occur when EJB 2.x interfaces are removed. Affected buckets include:
connector
javamail
jaxws
jdbc
jms
jta
samples
webservices12
xa

Note: Some tests will need additional updates to be fully compatible with EJB 3.x. This only takes care of the EJBVehicle itself.

I've done limited testing locally with OL to make sure that the general pattern used here is correct, but don't have any capacity to do a large scale rebuild of all of the affected components of the TCK.

In the event that this testing is ported to Arquillian, etc. these same updates will need to be made in the applications there in order to maintain compatibility with implementations that choose not to support the optional EJB 2.x interfaces.

Needed for https://github.com/jakartaee/platform-tck/issues/1250

@starksm64 @scottmarlow


scottmarlow (approved):

scottmarlow:
> In the event that this testing is ported to Arquillian, etc. these same updates will need to be made in the applications there in order to maintain compatibility with implementations that choose not to support the optional EJB 2.x interfaces.

Thanks for this note Brian, makes sense!

scottmarlow:
I'm thinking that we should merge your pull requests tomorrow.  If further changes are needed after merging, they can be made after.

scottmarlow:
Thanks Brian!

PR #1276, opened 2024-03-29
update parent pom relative path and add artifact installer for websocket tck.

gurunrao:
update parent pom relative path and add artifact installer for websocket tck

### DIFF
--- a/jpa/platform-tests/pom.xml
+++ b/jpa/platform-tests/pom.xml
@@ -23,7 +23,6 @@
         <groupId>jakarta.tck</groupId>
         <artifactId>persistence-tck</artifactId>
         <version>3.2.0</version>
-        <relativePath>../pom.xml</relativePath>
     </parent>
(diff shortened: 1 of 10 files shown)


jpa/platform-tests/pom.xml
```diff
@@ -23,14 +23,14 @@
...
         <version>3.2.0</version>
-        <relativePath>../pom.xml</relativePath>
     </parent>
     <artifactId>persistence-tck-platform-tests</artifactId>
     <packaging>jar</packaging>
     <dependencies>
         <dependency>
             <groupId>jakarta.servlet</groupId>
             <artifactId>jakarta.servlet-api</artifactId>
+            <version>6.0.0</version>
```
alwin-joseph, line 33: 6.1.0-M2 is the latest for EE1

gurunrao: 6.1.0 is the latest at the staging, hence added same.

PR #1284, opened 2024-04-12
Experimental support of GlassFish8 and JDK21 and JEE11 - do not merge!

dmatej:


scottmarlow:
@dmatej should we create a glassfish branch that only has things for running TCKs against GlassFish?  Most of our EE 11 TCK work is on the `tckrefactor` branch but not sure when that will be ready for platform testing.

dmatej:
My question is - when the `tckrefactor` branch will become a `master` branch? Currently it is rather a development branch which should be merged to `master`.

If you can manage to transform all tests to standalone, then we don't even need the `jakartaeetck.sh` bash script and `cts-base` image if I am right. But if you don't - this will still work (or should work) for "old" tests.

PR #1305, opened 2024-05-10
Fix EJB 2.x Refactor

brideck:
This supercedes Scott's change under https://github.com/jakartaee/platform-tck/pull/1304

There was code in the assembly suite that was also being used in the ejb suite. This is a problem because I had modernized the assembly suite to use EJB 3.x, but we explicitly want the ejb suite to remain at the EJB 2.x level. The same classes can't really work for both levels like was inadvertently being tried here.

I removed the dependency from the ejb suite, which involved bringing over 2.x compatible versions of all of the needed classes into the ejb suite.

All of this code has been compiled and built from scratch to the best of my ability, so it should be good to go. You can see that I even found an extraneous compilation problem in sec/mdb/MDBClient that I otherwise wouldn't have touched for this piece of work.

@starksm64 @scottmarlow


scottmarlow (approved):

scottmarlow:
Thanks Brian!

PR #1363, opened 2024-07-12
Fix poluted classpath and not testing what we thought we were

arjantijms:
Note to ourselves: we have several tests that are client-only, or have a client testing part (signature specifically). They way we setup these tests now makes them very vulnarable for classpath polution. We should probably start an explicit new process for this.

@starksm64 @scottmarlow

### DIFF
--- a/glassfish-runner/expression-language-tck/expression-language-tck-run/pom.xml
+++ b/glassfish-runner/expression-language-tck/expression-language-tck-run/pom.xml
@@ -20,14 +20,12 @@
 
     <modelVersion>4.0.0</modelVersion>
 
-    <parent>
-        <groupId>org.glassfish</groupId>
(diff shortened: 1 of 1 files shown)


dmatej (approved):

scottmarlow:
> Note to ourselves: we have several tests that are client-only, or have a client testing part (signature specifically). They way we setup these tests now makes them very vulnarable for classpath polution. We should probably start an explicit new process for this.

In general when you see `Java SE class path style testing` that is a sign of testing done before the EE platform implementation testing is ready to be run.  

Regardless, we should address when possible and note suggestions as well somewhere.  

I'm not really sure if we are on the same page but I was thinking that for EE 12, we should eliminate any testing done without EE deployment which impacts our current `Spec Wave` approach in a big way.

PR #1407, opened 2024-07-23
Restore the BaseUrlClient extending the EETest class and isolate the …

### LINKED ISSUE #1404: Changes to com.sun.ts.tests.common.webclient.BaseUrlClient have broken some ejb30 tests
When the servlet tests were ported Arquillian the com.sun.ts.tests.common.webclient.BaseUrlClient dropped its com.sun.ts.lib.harness.EETest base class in the e053c03a5fd8a066f5a6ec0cd54eec2f5e9e87a8 commit. This breaks the use of the vehicle framework by the javatest arquillian protocol in tests like com.sun.ts.tests.ejb30.misc.sameejbclass.Client.

We need to restore that behavior, but there are uses of this class in the jakarta-tags-tck and pages-platform-tck. It seems like the existing com.sun.ts.tests.common.webclient.BaseUrlClient should be copied into those tcks, and the common module version restored to the EETest based version.

starksm64:
…pages and tags tck usage.

Note that this is ignore the servlet module as I believe we are just deleting that code since it has moved to its own tck.

**Fixes Issue**
Fixes #1404 

**Related Issue(s)**
#1404 

**Describe the change**
This has restored the EETest base class to com.sun.ts.tests.common.webclient.BaseUrlClient and created separate versions without the EETest base class in the pages and tags tck modules.

@starksm64 @scottmarlow

### DIFF
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -104,5 +104,20 @@
                 </excludes>
             </resource>
         </resources>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-source-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>attach-sources</id>
+                        <goals>
+                            <goal>jar-no-fork</goal>
+                        </goals>
+                        <phase>verify</phase>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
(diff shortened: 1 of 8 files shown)


alwin-joseph (approved):
Tested jsp, jstl modules. LGTM

PR #1531, opened 2024-09-16
Create a release module for the tck distribution

### LINKED ISSUE #1523: Need to start a distribution module
We need to move away from using EE10 TCK contents for bootstrapping and have a EE 11 TCK dist that includes all ported test artifacts, the EE 11 ts.jte file, etc.

starksm64:
**Fixes Issue**
Fixes #1523

@starksm64 @scottmarlow


gurunrao (approved):
thanks for the PR @starksm64. LGTM.

el-tck/pom.xml
```diff
@@ -27,7 +27,7 @@
     </parent>
 
     <artifactId>jakarta-expression-language-tck</artifactId>
-    <version>6.0.1</version>
+    <version>11.0.0-SNAPSHOT</version>
```
alwin-joseph, line 30: el-tck module (artifactId: jakarta-expression-language-tck) is the standalone TCK which does not run in vehicles and is not part of platform TCK. The latest version released was 6.0.1 for the specification release.

el module is the right one which is refactored for platform TCK , and run in vehicles.

starksm64: Ok, this has been changed.

release/artifact-install.pom
```diff
@@ -0,0 +1,469 @@
...
+                    </execution>
+                    <execution>
+                        <id>install-el-tck</id>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>install-file</goal>
+                        </goals>
+                        <configuration>
+                            <groupId>jakarta.tck</groupId>
+                            <artifactId>el-tck</artifactId>
```
alwin-joseph, line 128: Should this be changed to "el-platform-tck" or rename el module artifact id to "el-tck"

alwin-joseph: I think it would be better to have a common naming convention for all modules in platform TCK. Maybe better to keep  "el-tck" and rename the artifact id in module el.

gurunrao: @alwin-joseph - Please note rename of EL standalone TCK, should not duplicate balloted EL standalone TCK artifact.That will impact future service releases of EL standalone TCK.

alwin-joseph: For standalone el-tck/ module, we have below :
```
<artifactId>jakarta-expression-language-tck</artifactId>
<name>el-tck</name>
```

For platform el/ module
```
<artifactId>el-platform-tck</artifactId>
<name>el</name>
``` 
Maybe we can continue to use `el-platform-tck` as artifact for platform(so as to not confuse with el-tck module `name` of the standalone) and correct the same here in line 128, for now.

PR #1536, opened 2024-09-18
Simplify glassfish runners for jaxrs/javaee modules, add Junit tags for jaxrs based on keyword.properties

alwin-joseph:
**Describe the change**
- Use glassfish.postBootCommands instead of maven plugin in javaee/jaxrs modules
- Add Junit tags for jaxrs tests based on keyword.properties


arjantijms:
Looks good!

Btw, if possible and where time allows (I'm very well aware we don't have much time) we should be using the right terms. JAX-RS is called simply REST in Jakarta EE 10 and Jakarta EE 11.

PR #1556, opened 2024-10-02
Updates for the Core Profile TCK

### LINKED ISSUE #1555: Core Profile: Rename the signature test file to exclude the `-jdk11` suffix
Currently the `core-profile-tck/ca-sigtest` signature test files includes a `sigtest-jdk11` Maven artifact qualifier which ends up being the file name suffix as well. The `-jdk11` part should be removed.

jamezp:
**Fixes Issue**
resolves #1555 

**Describe the change**
Upgrades the Jakarta Annotations API to 3.0 for the core-profile-tck. Also, renames the signature test file to remove the `-jdk11` suffix. 

Migrate the `sigtest-maven-plugin` to the latest vresion.

@starksm64 @scottmarlow

### DIFF
--- a/el-tck/docs/userguide/src/main/jbake/content/using-examples.inc
+++ b/el-tck/docs/userguide/src/main/jbake/content/using-examples.inc
@@ -22,9 +22,9 @@ sigtest-maven-plugin in the TCK runner as below.
 
 =======================================================================
 <plugin>
-    <groupId>org.netbeans.tools</groupId>
+    <groupId>jakarta.tck</groupId>
     <artifactId>sigtest-maven-plugin</artifactId>
-    <version>1.5</version>
+    <version>2.2</version>
     <configuration>
         <sigfile>path-to-sig-file-provided-with-TCK</sigfile>
         <packages>jakarta.el</packages>
--- a/core-profile-tck/ca-sigtest/pom.xml
+++ b/core-profile-tck/ca-sigtest/pom.xml
@@ -23,6 +23,11 @@
     <name>Common Annotations Signature Gen for Core Profile TCK</name>
     <description>This module contains is run to generate the signature test file</description>
 
+    <properties>
+        <sigtest.file.name>common-annotations-api.sig</sigtest.file.name>
+        <sigtest.path>${project.basedir}/src/main/resources/${sigtest.file.name}</sigtest.path>
+    </properties>
+
     <dependencies>
         <dependency>
             <groupId>jakarta.annotation</groupId>
@@ -47,9 +52,9 @@
                         <configuration>
                             <artifacts>
                                 <artifact>
-                                    <file>src/main/resources/common-annotations-api-jdk11.sig</file>
+                                    <file>${sigtest.path}</file>
                                     <type>sig</type>
-                                    <classifier>sigtest-jdk11</classifier>
+                                    <classifier>sigtest</classifier>
                                 </artifact>
                             </artifacts>
                         </configuration>
@@ -102,12 +107,13 @@
                         <groupId>jakarta.tck</groupId>
                         <artifactId>sigtest-maven-plugin</artifactId>
                         <configuration>
-                            <classes>${project.build.directory}/ca-api</classes>
+                            <classes>${project.build.directory}/ca-api/</classes>
                             <packages>jakarta.annotation,
                 jakarta.annotation.security,
                 jakarta.annotation.sql</packages>
                             <attach>false</attach>
-                            <sigfile>${project.build.directory}/common-annotations-api-jdk11.sig</sigfile>
+                            <sigfile>${sigtest.path}</sigfile>
+                            <release>${maven.compiler.release}</release>
                         </configuration>
                         <executions>
                             <execution>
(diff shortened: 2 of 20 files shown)


jamezp:
There is nothing I can do about the build failure since the build requires a SNAPSHOT build of https://github.com/eclipse-ee4j/jakartaee-tck-tools/tree/master/arquillian

el-tck/docs/userguide/src/main/jbake/content/using-examples.inc
```diff
@@ -22,9 +22,9 @@ sigtest-maven-plugin in the TCK runner as below.
 
 =======================================================================
 <plugin>
-    <groupId>org.netbeans.tools</groupId>
+    <groupId>jakarta.tck</groupId>
     <artifactId>sigtest-maven-plugin</artifactId>
-    <version>1.5</version>
+    <version>2.2</version>
```
scottmarlow, line 27: 2.3 is probably a better release as it has better back compatibility.

jamezp: The issue with 2.3 is it breaks the `-IgnoreJDKClass` argument which _some_ signature test runners use. With that change you can only pass one `-IgnoreJDKClass` argument, which causes signature tests that pass it more than once to fail. In previous versions passing it for each type you want to ignore was the way to do it.

I think https://github.com/eclipse-ee4j/jakartaee-tck-tools/pull/40 somewhat undoes it, but IIRC it doesn't allow passing parameters to `-IgnoreJDKClass` which also breaks compatibility. 

It feels safer to stick with 2.2 for now.

gurunrao:
Annotation TCK source has been moved to https://github.com/jakartaee/common-annotations-api/tree/master/tck.
IMO, core-profile should download and run Annotation TCK for EE 11, other option is to add documentation in core-profile for user to run annotation TCK.
I believe duplication of signatures in multiple github repository will cause issues for future maintenance/releases.

CC @scottmarlow @scottkurz

scottmarlow:
> Annotation TCK source has been moved to https://github.com/jakartaee/common-annotations-api/tree/master/tck. IMO, core-profile should download and run Annotation TCK for EE 11, other option is to add documentation in core-profile for user to run annotation TCK. I believe duplication of signatures in multiple github repository will cause issues for future maintenance/releases.

I think that the underlying question is should someone running the Core Profile TCK which will include signature tests for all specs included in Core Profile have to run separate signature TCK tests in Annotations/other SPEC TCKs?

The same question for Platform/Web Profile TCK, should users have to run signature tests when running other SPEC TCKs?

gurunrao:
> > Annotation TCK source has been moved to https://github.com/jakartaee/common-annotations-api/tree/master/tck. IMO, core-profile should download and run Annotation TCK for EE 11, other option is to add documentation in core-profile for user to run annotation TCK. I believe duplication of signatures in multiple github repository will cause issues for future maintenance/releases.
> 
> I think that the underlying question is should someone running the Core Profile TCK which will include signature tests for all specs included in Core Profile have to run separate signature TCK tests in Annotations/other SPEC TCKs?
> 
> The same question for Platform/Web Profile TCK, should users have to run signature tests when running other SPEC TCKs?

Annotation TCK has an Signature Test, no other tests are run as part of TCK. Hence my point of non duplication of Signature file [ since the file has already been moved to https://github.com/jakartaee/common-annotations-api/tree/master/tck ].

scottmarlow:
> > > Annotation TCK source has been moved to https://github.com/jakartaee/common-annotations-api/tree/master/tck. IMO, core-profile should download and run Annotation TCK for EE 11, other option is to add documentation in core-profile for user to run annotation TCK. I believe duplication of signatures in multiple github repository will cause issues for future maintenance/releases.
> > 
> > 
> > I think that the underlying question is should someone running the Core Profile TCK which will include signature tests for all specs included in Core Profile have to run separate signature TCK tests in Annotations/other SPEC TCKs?
> > The same question for Platform/Web Profile TCK, should users have to run signature tests when running other SPEC TCKs?
> 
> Annotation TCK has an Signature Test, no other tests are run as part of TCK. Hence my point of non duplication of Signature file [ since the file has already been moved to https://github.com/jakartaee/common-annotations-api/tree/master/tck ].

I agree now  and understand the point better.  

So the Core Profile TCK is currently delegating to the other Specification TCKs to cover signature testing, so  we should do the same with the Annotation TCK which does now cover the signature testing as well.

scottmarlow:
What do you think @jamezp? about ^

scottmarlow:
Just to add more info, for Platform/Web Profile TCKs we always check every SPEC API signature included by the Platform/Web Specification instead of delegating that to the Spec implementation TCKs.

jamezp:
I can understand the desire to not duplicate efforts. I don't really have a strong opinion either way. The one thing we'd need to think about though is how we run it from the Core Profile TCK. While it should be fairly easy from a TCK runner, there's no way to validate they're running the correct TCK. That's the one part I see value in having it be part of the Core Profile TCK.

starksm64:
The user guide is actually a bit opaque on how the common annotations signature tests are run. I guess you would only know to use the core-profile-tck/ca-sigtest by looking at the wf-core-tck-runner example as the user guide itself just says it needs to be run, but does not mention it in the section `Running the Signature Tests` section. It should be added there, and we can just delegate to how to run it to the component spec tck as we do the other component specs in the core profile.

I'm just going to merge this and do some updates the to user guide and dist.

PR #1561, opened 2024-10-04
Refactor JPA tests (work in progress)

scottmarlow:
**Fixes Issue**
Work in progress for https://github.com/jakartaee/platform-tck/issues/1371

**Describe the change**
3578 of the PmServlet/PuServlet tests are passing.  Still have some failures to work out in PmServlet/PuServlet tests.  Likely more setup/cleanup issues.

just made a change for appclient  tests which is where most of the remaining work is but that might just require syncing the PmServlet/PuServlet changes over mostly, will see.

@starksm64 @scottmarlow


common/src/main/java/com/sun/ts/lib/harness/ServiceEETest.java
```diff
@@ -94,7 +94,7 @@ public Status run(String[] argv, Properties p) {
       String className = this.getClass().getName();
       // use this name for the context root or jndi name to eliminate
       // naming conflicts for apps deployed at the same time
-      String sVehicleEarName = TestUtil.getProperty(p, "vehicle_ear_name");
+      // comment out unused: String sVehicleEarName = TestUtil.getProperty(p, "vehicle_ear_name");
```
scottmarlow, line 97: We could leave this in or remove it.  If anyone finds they need the vehicle_ear_name property, we should add it back in.

scottmarlow:
Dear reviewers, please review the minor change to /common source.  The jpa changes are still working in progress but good progress so far.  I think that remaining work for appclient is mostly passing the tests that we already are passing (so mostly test changes needed to deal with setup/clean type logic).  

I think we either need to execute the ddl/dml create scripts more often or we need to do less clean up as one problem with a few tests seemed to be caused by deleting the tables created by the /sql/derby scripts too soon (basically I could see a failure due to the table being missing that would of been created in those scripts).  This is all minor at this point.  

Just to share it here in case this is helpful for other changes, I fixed a deployment problem with the jpa appclient tests today by running script (see the commit from today with output):

```
function update_tests ()
{
  name=$1

    for updatefile in `find  -name "*$name*"`; do
        # echo "do something with file $updatefile"
        EAR_CLASS="$(grep "EnterpriseArchive.class" $updatefile | cut -d \" -f2 | cut -d'.' -f1)"
        sed "s%JavaArchive.class.*_client.jar%JavaArchive.class, \"${EAR_CLASS}_client.jar%" -i $updatefile
    done
}

# covers both AppmanagedTest + Appmanagednotx
update_tests "*Appmanaged*.java"
update_tests "*Stateful3Test*.java"
update_tests "*Stateless3Test*.java"
```

PR #1573, opened 2024-10-09
The spotless plugin seems to mess up the Maven reactor when sortModul…

jamezp:
…es is set to true. This is okay for the check goal, but not apply. Add the configuration to the check executions and remove sortModules from the default configuration.

@starksm64 @scottmarlow


jamezp:
The simple option to fix this is to simply just remove the `<sortModules>true</sortModules>` all together. However, if we want to validate they are in alphabetical order, this is the way to do that.

starksm64 (approved):

PR #1647, opened 2024-11-12
arquillian/appclient wip progress logging changes

scottmarlow:
Make the appclient logging a little less verbose by default at line 86.  

I'm not yet sure why the logged information at 122 doesn't show the ear deployment file name (instead it seems to show the client jar (which is part of the ear).  It would be nice to show the top level deployment as well.

### DIFF
--- a/arquillian/appclient/src/main/java/tck/arquillian/protocol/appclient/AppClientMethodExecutor.java
+++ b/arquillian/appclient/src/main/java/tck/arquillian/protocol/appclient/AppClientMethodExecutor.java
@@ -61,6 +61,8 @@ public class AppClientMethodExecutor implements ContainerMethodExecutor {
     @Override
     public TestResult invoke(TestMethodExecutor testMethodExecutor) {
         TestResult result = TestResult.passed();
+        String appArchiveName = null;
+        String vehicleArchiveName = null;
 
         // Run the appclient for the test if required
         String testMethod = testMethodExecutor.getMethodName();
@@ -68,8 +70,8 @@ public class AppClientMethodExecutor implements ContainerMethodExecutor {
             log.info("Running appClient for: " + testMethod);
             try {
                 Deployment deployment = deploymentInstance.get();
-                String appArchiveName = appClientArchiveName.get().name();
-                String vehicleArchiveName = TsTestPropsBuilder.vehicleArchiveName(deployment);
+                appArchiveName = appClientArchiveName.get().name();
+                vehicleArchiveName = TsTestPropsBuilder.vehicleArchiveName(deployment);
                 String[] additionalAgrs = TsTestPropsBuilder.runArgs(config, deployment, testMethodExecutor);
                 appClient.run(vehicleArchiveName, appArchiveName, additionalAgrs);
             } catch (Exception ex) {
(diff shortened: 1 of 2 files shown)


scottmarlow:
CC @jamezp

arquillian/appclient/src/main/java/tck/arquillian/protocol/appclient/AppClientMethodExecutor.java
```diff
@@ -81,13 +83,13 @@ public TestResult invoke(TestMethodExecutor testMethodExecutor) {
...
 
-        log.info(String.format("AppClient(%s) readAll returned %d lines\n", testMethod, lines.length));
+        log.fine(String.format("AppClient(%s) readAll returned %d lines\n", testMethod, lines.length));
         boolean sawStatus = false;
         MainStatus status = MainStatus.NOT_RUN;
         String reason = "None";
         String description = "None";
         for (String line : lines) {
-            System.out.println(line);
+            log.finer(line);
```
jamezp, line 92: This part could end up making it look a little odd if it's already a formatted log line. I also feel this should be at level `fine` not `finer`.

arquillian/appclient/src/main/java/tck/arquillian/protocol/appclient/AppClientMethodExecutor.java
```diff
@@ -117,6 +119,7 @@ public TestResult invoke(TestMethodExecutor testMethodExecutor) {
                     break;
             }
             result.addDescription(description);
+            log.info(result.toString() + ": " + description + ": " + vehicleArchiveName + ": " + appArchiveName);
```
jamezp, line 122: I think this might look a little better with something like:
```
log.log(level, String.format("%s_%s [%s] %s : %s", vehicaleArchiveName, appArchiveName, result.getStatus(), description, result);
```

Note I'm not sure how the `vehicaleArchiveName` and `appArchiveName` should be formatted.

The `level` should be set in the in the switch statement right above this log line IMO. Then for `ERROR` and `FAILED` status we could set the level to `Level.SEVERE`.

jamezp (commented):
Overall this seems reasonable. One question I'd ask though is will we always be running with DEBUG logs enabled? Given what is logged, I assume yes. If so, we might want to consider keeping them at `INFO`, but I don't have a strong opinion there.

arquillian/appclient/src/main/java/tck/arquillian/protocol/appclient/AppClientMethodExecutor.java
scottmarlow: Thanks, switching over to this style.  I don't think the appclient jar or vehicle names are what I really wanted to show here.  Showing the EAR name would be nicer (based on how we current include more information in the ear name).  Revising...

scottmarlow:
> Overall this seems reasonable. One question I'd ask though is will we always be running with DEBUG logs enabled? Given what is logged, I assume yes. If so, we might want to consider keeping them at INFO, but I don't have a strong opinion there.

I have to verify as currently, I have the platformtck/glassfish-runner/persistence-platform-tck/jakartaeetck/bin/ts.jte `harness.log.traceflag=true` which might influence some logging levels.

arquillian/appclient/src/main/java/tck/arquillian/protocol/appclient/AppClientMethodExecutor.java
scottmarlow: Will update this to show the new output with:
```
            result.addDescription(description);
            log.log(Level.INFO, String.format("[%s] %s : %s", result.getStatus(), description, result));

```

arquillian/appclient/src/main/java/tck/arquillian/protocol/appclient/AppClientMethodExecutor.java
```diff
@@ -117,6 +120,7 @@ public TestResult invoke(TestMethodExecutor testMethodExecutor) {
                     break;
             }
             result.addDescription(description);
+            log.log(Level.FINE, String.format("[%s] %s : %s", result.getStatus(), description, result));
```
jamezp, line 123: I still feel like it makes sense to make the level configurable. Meaning, use a level based on the status.

arjantijms:
@scottmarlow is this PR still relevant?

scottmarlow:
> @scottmarlow is this PR still relevant?

Probably not anymore.

PR #1729, opened 2025-01-02
Prepare tags TCK for standalone release

arjantijms:
@starksm64 @scottmarlow

### DIFF
--- a/tcks/apis/tags/.gitignore
+++ b/tcks/apis/tags/.gitignore
@@ -1,0 +1,1 @@
+/.flattened-pom.xml
--- a/tcks/apis/tags/pom.xml
+++ b/tcks/apis/tags/pom.xml
@@ -20,10 +20,10 @@
     <modelVersion>4.0.0</modelVersion>
 
     <parent>
-        <groupId>jakarta.tck</groupId>
+        <groupId>org.eclipse.ee4j</groupId>
         <artifactId>project</artifactId>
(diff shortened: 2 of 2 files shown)


scottmarlow:
Which EE release is this for?  Something after EE 11 like 12?

arjantijms:
Don't we run the (refactored) pages TCK for ratification of the EE 11 platform?

Within the refactored TCKs we have a numbers of TCKs that already passed for the individual components (such as expression language, persistence, etc), and I thought we would all be running these again for EE 11 platform.

Regardless, the refactored TCK will be the base for ratification of the Pages component spec in EE 12 (should Pages be updated for EE 12, of course)

scottmarlow:
> Don't we run the (refactored) pages TCK for ratification of the EE 11 platform?

Yes for Pages TCK just to make sure that the Pages implementation still passes.  For Tags we could do the same but I think with the https://www.eclipse.org/downloads/download.php?file=/jakartaee/tags/3.0/jakarta-tags-tck-3.0.1.zip

> 
> Within the refactored TCKs we have a number of TCKs that already passed for the individual components (such as expression language, persistence, etc), and I thought we would all be running these again for the EE 11 platform.

Yes, we can run the Standalone TCKs again but we need to ensure that we use the right version of the standalone TCKs.

> 
> Regardless, the refactored TCK will be the base for ratification of the Pages component spec in EE 12 (should Pages be updated for EE 12, of course)

+1

PR #1820, opened 2025-01-28
Change the test order in failing persistence tests

alwin-joseph:
**Describe the change**
- Changing the order of the tests passes all the remaining 6 failing tests in persistence web profile.
- This can be considered as a temporary fix to understand underlying issue. Possibly we should fix without relying on the order of the tests IMO. I am trying to find the cleanup/setup required to pass them otherwise. Please suggest if we can consider this as the permanent solution.

CC @arjantijms @gurunrao 
@starksm64 @scottmarlow


scottmarlow:
> Please suggest if we can consider this as the permanent solution.

Yes, Ordering of the tests is fine.  An improvement can be made via a different change if anyone wants to make a further change (e.g. remove the order and add any needed cleanup/setup changes).

arjantijms:
Wow, amazing! Very good work @alwin-joseph! I also took a look at these tests, but did not see that the order was the problem.

I think we can then set this PR to ready to review, or do you have other changes that need to be done?

PR #1873, opened 2025-02-04
Fix the build

arjantijms:
The build now fails on every PR check. This was caused by a plugin called spotless.

Removing this plugin also simplifies the build in other placed, where can now build again with regular mvn clean install instead of mvn spotless:apply && mvn clean install.

@starksm64 @scottmarlow

### DIFF
--- a/arquillian/pom.xml
+++ b/arquillian/pom.xml
@@ -85,16 +85,4 @@
             <version>5.11.4</version>
         </dependency>
     </dependencies>
-
-    <build>
-        <plugins>
-            <plugin>
-                <groupId>com.diffplug.spotless</groupId>
-                <artifactId>spotless-maven-plugin</artifactId>
-                <configuration>
-                    <skip>true</skip>
-                </configuration>
-            </plugin>
-        </plugins>
-    </build>
 </project>
--- a/glassfish-runner/enterprise-beans-tck/enterprise-beans-tck-run/pom.xml
+++ b/glassfish-runner/enterprise-beans-tck/enterprise-beans-tck-run/pom.xml
@@ -12,11 +12,11 @@
   
   ejb30 lite env tests:
   
-  mvn spotless:apply && mvn clean install -Dglassfish.version=8.0.0-SNAPSHOT -Dmaven.build.cache.enabled=false -Pejb30,exclude-timeout,ejb30_lite_env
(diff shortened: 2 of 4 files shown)


scottmarlow:
Thanks Arjan!  I was thinking about doing this but appreciate that you took action!

olamy:
we do not have anymore pom format check? This might turn into some mess... :(

olamy:
not sure to understand why this was required
```
where can now build again with regular mvn clean install instead of mvn spotless:apply && mvn clean install.
```
There might be some other issues here hidden.
Because once the pom is correctly formatted and committed you do not need to `spotless:apply` each time...

scottmarlow:
Hi Olivier,

We will enable spotless again when possible but right now it was the
easiest change to fix the Platform TCK build failures.  Spotless has served
us well and there is a price to disabling it but we are trying to complete
the EE 11 release (really just the Web Profile part first and then the
Platform).

Scott

starksm64:
Find a format plugin that actually tells you what the problem is.
Spitting out a chunk of pom and saying there is a problem without the
exact location and details is a poor ux.

olamy:
do you have some logs where this was failing? Because I cannot reproduce locally.
Maybe I'm not using correct profiles or option you are using.
But I will be happy to fix this. 
usually the location is easy to understand from the output. 
Anyway And you don;t even need to really read but just apply the auto formatting.

starksm64:
But that is the other problem. We were running into situations where
running 'mvn spotless:apply' simply failed with some error about not
finding the plugin.

olamy:
I perfectly understand the frustration.
However, such an error is due to the project's current poms hierarchy.
The plugin was declared in the top pom.
But child projects don't have this top pom as a parent so plugin declaration is not inherited but child projects are executed as declared in the `<modules>` which makes the hierarchy/inheritance totally inefficient,
such this 
```
   <parent>
        <groupId>org.eclipse.ee4j</groupId>
        <artifactId>project</artifactId>
        <version>1.0.9</version>
        <relativePath />
    </parent>
    
    <groupId>jakarta.tck</groupId>
    <artifactId>glassfish.annotations-tck</artifactId>
    <version>11.0.0-SNAPSHOT</version>
```
so when running the module Maven doesn't have any idea about spotless plugin as it's not part of `org.eclipse.ee4j:project:1.0.9`

Is there any particular reason to not have a standard parent/child structure that would make it easier to maintain poms and remove all the current duplication?

PR #1900, opened 2025-02-08
issue/1895 use jakarta.jakartaee-bom

scottmarlow:
**Fixes Issue**
Help with https://github.com/jakartaee/platform-tck/issues/1895 but needs further change as some sub-modules don't use the root project versions from the bom it seems

@starksm64 @scottmarlow

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -79,53 +79,10 @@
         <arquillian.container.se.api.version>1.0.2.Final</arquillian.container.se.api.version>
         <arquillian.version>1.9.1.Final</arquillian.version>
 
-        <!-- CDI -->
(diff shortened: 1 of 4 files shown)


starksm64:
What you need to do is to remove the actual component api dependencies from the  dependencyManagement of the jakarta.tck:project pom.

So for example, remove this:
```xml
        <dependency>
            <groupId>jakarta.jms</groupId>
            <artifactId>jakarta.jms-api</artifactId>
        </dependency>
```

Then this dependency in the tcks/apis/messaging/messaging-inside-container/pom.xml does point to the dependency from the jakarta api bom.

PR #1939, opened 2025-02-17
add web junit tag to jdbc tests.

gurunrao:
add web junit tag to jdbc tests.


alwin-joseph:
Maybe we can add "platform" tag also to all tests to be in sync with other modules.

arjantijms (approved):

PR #1974, opened 2025-02-23
Fix resource inclusion in TCK jars

arjantijms:
Because of including the license file, the default resource inclusion was broken. All jars files for TCKs that needed resource inclusion were broken.

This PR fixes that.

@starksm64 @scottmarlow

### DIFF
--- a/tcks/apis/cdi-ee-tck/tck/pom.xml
+++ b/tcks/apis/cdi-ee-tck/tck/pom.xml
@@ -388,6 +388,12 @@
                 </includes>
                 <targetPath>META-INF</targetPath>
             </resource>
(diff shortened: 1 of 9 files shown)


edburns:
I take responsibility for this. My changes for #1955 caused this regression in the case where a POM did not have a `<resources>` element. I suspect by adding a `<resources>` element where none had been caused the implicit rule to include `src/main/resources` to no longer take effect. Sorry about that, and thanks @arjantijms  for fixing it.

arjantijms:
No problem, exactly this occured to me many times before.

A good lesson is that the CI needs to run the latest version of what we stage.

Issue #1991, opened 2025-02-26
Need final releases of tools and common


starksm64:
**Is your feature request related to a problem? Please describe.**
We have dependencies for the platform tck that are being released independently. These need to be updated to a final release and the associated web profile dependencies updated. Right now there is inconsistent references to these artifact versions.

**Describe the solution you'd like**
We need a single place to set these dependency versions.

PR #2001, opened 2025-02-27
Cleanup versions by introducing consistent version properties

starksm64:
Add jakarta.tck.sigtest.version property for jakarta.tck:signaturetest
Change jakarta.tck.tools.version to jakarta.tck.common.version
Use jakarta.jakartaee-bom.version for the jakarta.platform:jakarta.jakartaee-bom version
Use jakarta.tck.arquillian.version for jakarta.tck.arquillian:* artifacts

**Related Issue(s)**
Related to #1991


bom/pom.xml
```diff
@@ -33,8 +33,8 @@
   <name>Jakarta EE TCK Artifacts BOM</name>
 
   <properties>
-    <version.jakarta.tck.arquillian>11.0.0-RC1</version.jakarta.tck.arquillian>
-    <version.jakarta.tck.common>11.0.0-RC5</version.jakarta.tck.common>
+    <version.jakarta.tck.arquillian>11.0.0-RC2</version.jakarta.tck.arquillian>
```
jamezp, line 36: We should probably rename this property to `jakarta.tck.arquillian.version` to be consistent with the others.

jamezp (approved):
I added one minor comment. This looks good though.

Somewhat of a potentially loaded question, should we consider having a BOM to manage the versions? I ask because they seem all over the place which is difficult to manage.

starksm64:
> Somewhat of a potentially loaded question, should we consider having a BOM to manage the versions? I ask because they seem all over the place which is difficult to manage.

Yes, this would be ideal and what we are working towards with the jakarta.tck:artifacts-bom

PR #2003, opened 2025-02-27
Alternative approach to simply version inconsistencies 

starksm64:
Alternative approach to simply version inconsistencies using the jakarta.tck:artifacts-bom to eliminate subproject pom version properties.

**Related Issue(s)**
Related to #1991

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -21,6 +21,7 @@
     <groupId>org.eclipse.ee4j</groupId>
     <artifactId>project</artifactId>
     <version>1.0.9</version>
+    <relativePath />
   </parent>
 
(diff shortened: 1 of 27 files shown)


jamezp (approved):
This looks good to me.

I can see needing, not in this PR as it's a different issue, a `jakarta.tck:jakarta-platform-bom`, `jakarta.tck:jakarta-web-bom` and `jakarta.tck:jakarta-core-bom`. If that seems reasonable, I can file a different issue and file a PR after this is merged.

starksm64:
Yes, profile boms would be good as I did not update the core profile, and trimming the artifacts for a web profile run would be good.

PR #2026, opened 2025-03-05
Signature test fixes

### LINKED ISSUE #2024: Remove the jimage invocation for signature tests
The `com.sun.ts.tests.signaturetest.SigTestEE` invokes the `jimage extract` command to extracts jimage entries from he current JVM. This is no longer needed and we shouldn't need to extract any types to the file system.

As part of this we should deprecate the `SigTestData.getJImageDir()` as it's no longer used.

### LINKED ISSUE #2025: Remove the -IgnoreJDKClass individual arguments in favor of -IgnoreJDKClasses to ignore them all and allow the -BootCp to be passed
We currently have sever JDK classes excluded with the `-IgnoreJDKClass` argument. We should change to use the `-IgnoreJDKClasses` and ignore them all. This helps reflection tests which always load the description based on the currently running JVM.

We should also add the `-BootCp 17` argument to indicate we're testing with Java 17. We should also consider making this configurable. This setting helps static tests where we use the `ct.sym` file for the specific version from the JDK to match signatures.

jamezp:
**Fixes Issue**
* resolves #2024
* resolves #2025

**Describe the change**

This removes the `jimage` command from being executed. It also replaces the individual `-IgnoreJDKClass` arguments with the single `-IgnoreJDKClasses`. 

Add the `-BootCp` property to the command with the ability to override it, while defaulting to 17. This is the minimum version which the signature test files are generated to.

Update the `JAVA_FULL` value from `full` to `platform` to match the JUnit 5 tag. Update the configuration documentation to include some additional signature test properties.

Upgrade the sigtest-maven-plugin to 2.6 and include a missing dependency.

@starksm64 @scottmarlow

### DIFF
--- a/tcks/profiles/platform/docs/userguide/platform/src/main/asciidoc/config.adoc
+++ b/tcks/profiles/platform/docs/userguide/platform/src/main/asciidoc/config.adoc
@@ -340,6 +340,7 @@ These are the properties that need to have a value in the `ts.jte` file provided
 * imap.port
 * iofile
 * java.naming.factory.initial
+* javaee.level
 * javamail.mailbox
 * javamail.password
 * javamail.protocol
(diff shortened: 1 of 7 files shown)


scottmarlow:
Thanks James!

PR #2042, opened 2025-03-06
Update jakarta.tck:common to 11.0.0-RC7

### LINKED ISSUE #2039: Update jakarta.tck:common to 11.0.0-RC7
Update jakarta.tck:common jakarta.tck.common.version property to 11.0.0-RC7 to validate the candidate final release version

starksm64:
**Fixes Issue**
Fixes #2039

### DIFF
--- a/tools/common/pom.xml
+++ b/tools/common/pom.xml
@@ -29,7 +29,7 @@
 
     <groupId>jakarta.tck</groupId>
     <artifactId>common</artifactId>
-    <version>11.0.0-RC7</version>
+    <version>11.0.0-SNAPSHOT</version>
     <packaging>jar</packaging>
 
     <name>TCK Tools Common</name>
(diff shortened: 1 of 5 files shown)


tools/common/pom.xml
```diff
@@ -29,7 +29,7 @@
 
     <groupId>jakarta.tck</groupId>
     <artifactId>common</artifactId>
-    <version>11.0.0-SNAPSHOT</version>
+    <version>11.0.0-RC7</version>
```
jamezp, line 32: Is this part right? I wouldn't think we'd want this pushed to main because every time it's built locally it will replace what has, potentially, been downloaded from Maven.

starksm64: Correct, that is not right. Fixed.

jamezp (approved):

Issue #2064, opened 2025-03-12
Standalone TCK artifacts have been pushed to central


starksm64:
**Describe the bug**
We cannot push the current staged TCK to central because of the tcks/apis/websocket has a fixed 2.2.0 version that has already been pushed to maven central as part of previous 11.0.0-M? releases.

**To Reproduce**
```
[ERROR] 
[ERROR] Nexus Staging Rules Failure Report
[ERROR] ==================================
[ERROR] 
[ERROR] Repository "jakartatck-1206" failures
[ERROR]   Rule "RepositoryWritePolicy" failures
[ERROR]     * Artifact updating: Repository ='releases:Releases' does not allow updating artifact='/jakarta/tck/transaction-docs/2.0/transaction-docs-2.0.pom'
[ERROR]     * Artifact updating: Repository ='releases:Releases' does not allow updating artifact='/jakarta/tck/transaction-docs/2.0/transaction-docs-2.0-cyclonedx.xml'
[ERROR]     * Artifact updating: Repository ='releases:Releases' does not allow updating artifact='/jakarta/tck/transaction-docs/2.0/transaction-docs-2.0-cyclonedx.json'
[ERROR]     * Artifact updating: Repository ='releases:Releases' does not allow updating artifact='/jakarta/tck/connector-docs/2.1/connector-docs-2.1.pom'
[ERROR]     * Artifact updating: Repository ='releases:Releases' does not allow updating artifact='/jakarta/tck/connector-docs/2.1/connector-docs-2.1-cyclonedx.json'
[ERROR]     * Artifact updating: Repository ='releases:Releases' does not allow updating artifact='/jakarta/tck/connector-docs/2.1/connector-docs-2.1-cyclonedx.xml'
[ERROR]     * Artifact updating: Repository ='releases:Releases' does not allow updating artifact='/jakarta/tck/websocket-tck-spec-tests/2.2.0/websocket-tck-spec-tests-2.2.0-sources.jar'
[ERROR]     * Artifact updating: Repository ='releases:Releases' does not allow updating artifact='/jakarta/tck/websocket-tck-spec-tests/2.2.0/websocket-tck-spec-tests-2.2.0-cyclonedx.json'
[ERROR]     * Artifact updating: Repository ='releases:Releases' does not allow updating artifact='/jakarta/tck/websocket-tck-spec-tests/2.2.0/websocket-tck-spec-tests-2.2.0-cyclonedx.xml'
[...]

PR #2073, opened 2025-03-12
Skip the nexus-staging-maven-plugin for the fixed version standalone …

starksm64:
…tck docs

**Fixes Issue**
#2064

### DIFF
--- a/tcks/apis/connector/docs/pom.xml
+++ b/tcks/apis/connector/docs/pom.xml
@@ -33,21 +33,13 @@
     <name>connector tck</name>
     <description>Connector docs</description>
 
+    <properties>
+        <maven.deploy.skip>true</maven.deploy.skip>
+    </properties>
+
     <modules>
         <module>parent</module>
         <module>userguide</module>
     </modules>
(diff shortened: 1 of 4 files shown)


tcks/apis/connector/docs/pom.xml
```diff
@@ -33,21 +33,13 @@
     <name>connector tck</name>
     <description>Connector docs</description>
 
+    <properties>
+        <maven.deploy.skip>true</maven.deploy.skip>
```
jamezp, line 37: We might also need `<skipNexusStagingDeployMojo>true</skipNexusStagingDeployMojo>` according to https://github.com/sonatype/nexus-maven-plugins/blob/main/staging/maven-plugin/README.md#plugin-flags.

starksm64: I tested running the nexus-staging:deploy locally and it does skip these, so I thin the current setting in the parent pom that reads the maven.deploy.skip property as the default skipNexusStagingDeployMojo is working.

PR #2078, opened 2025-03-13
issues/2074 The appclient tests should be updated to include appropri…

scottmarlow:
…ate @tag but change is not needed for Web Profile

**Fixes Issue**
Specify the issue (link) that is solved with this pull request.

**Related Issue(s)**

**Describe the change**
A clear and concise description of the change.

**Additional context**

@starksm64 @scottmarlow


starksm64:
These appclient tests still need to be fully ported over to Arquillian as they have no deployment methods yet. At that time whey will also need to have the `@Tag("tck-appclient")` and some other changes for the vehicle, etc.

arjantijms:
I'm still not convinced we should spend much (or any) time on appclient tests. Maybe we should bring this to the mailing list and the working group members again.

With the last vote about dropping appclient, I couldn't escape the feeling the ask was ill understood and that people voting thought we'd finish the entire TCK with appclient maybe 2 weeks later at most. It's now almost April 2025, and we haven't even released the web platform formally yet.

If we continue down this path I really think my previously estimated September 2025 release date is not so outragious. Of course I still hope to be proven wrong.

PR #2187, opened 2025-04-11
Address ee.jakarta.tck.persistence.core.annotations.version.Client* failures by allowing null to be used for initial version field/property value 

scottmarlow:
**Fixes Issue**
https://github.com/jakartaee/platform-tck/issues/2112

**Related Issue(s)**

**Describe the change**
EclipseLink version support can use Null for the initial version for certain cases.  This change is to allow null to be used for those cases.

@starksm64 @scottmarlow


scottmarlow:
hmm, will definitely remove the pom.xml change :)

arjantijms:
Same comment as: https://github.com/jakartaee/platform-tck/pull/2184#issuecomment-2797170461

If we make changes here, do we need to sync those to the original that we're going to transfer?

PR #2374, opened 2025-06-13
Added support for PKCS12 keystores, removed NSS, keeping JKS

dmatej:
- Adds support for future GlassFish 7.1.x without abandoning 7.0.x (which will continue to be used to validate TCK10 versions).
- Removed certutil usages - that doesn't exist even in 7.0.x, this was already unusable.
- Removed support for NSS - I remember that was supported in SGES 2, I did not use it since then as the standard JKS improved in later Java versions (6+)
- Added support for PKCS12, current industry standard.
  - This will be also probably used in future GlassFish versions starting with 7.1.0
- Added --add-opens for named "jdke" module

@starksm64 @scottmarlow


dmatej:
Status: testing against https://github.com/eclipse-ee4j/glassfish/pull/25517 . 
What I want is that it would work with
* GF 7.0.0-SNAPSHOT (nightly) and JDK11 as it is used to automatically validate PRs here.
* GF 7.1.0-SNAPSHOT and JDK17 here in TCK CI, using the same job
* GF 7.1.0-SNAPSHOT integrated with GlassFish's POM (doesn't use the cts docker image, uses mail server in docker image, TckRunner java class) locally AND in my Jenkins.

Simply said, it must be robust, stable and at reasonably flexible.

arjantijms (approved):

PR #2322, opened 2025-05-19
Defend against intermittent EJB failures

brideck:
1) Timer tests using 1-second timers will sometimes fail with 'Timer does not exist' failures. We've had bugs opened in the past for these, and the solution there has been to bump these timers to use a 2 second duration so they don't expire before the test is able to perform its necessary test logic. This updates a couple cases that have been missed.

2) Make local copy of Collections to avoid ConcurrentModificationException
e.g.
java.util.ConcurrentModificationException
	at java.base/java.util.ArrayList$Itr.checkForComodification(ArrayList.java:1013)
	at java.base/java.util.ArrayList$Itr.next(ArrayList.java:967)
	at java.base/java.util.AbstractCollection.toString(AbstractCollection.java:456)
	at java.base/java.util.Collections$UnmodifiableCollection.toString(Collections.java:1047)
	at java.base/java.lang.String.valueOf(String.java:4992)
	at java.base/java.lang.StringBuilder.append(StringBuilder.java:173)
	at com.sun.ts.tests.ejb30.common.lite.EJBLiteJsfClientBase.appendReason(EJBLiteJsfClientBase.java:382)
        ...

Both changes have been tested against Open Liberty and have been deemed non-disruptive.

CC @alwin-joseph @anajosep @arjantijms @cesarhernandezgt @dblevins @m0mus @edbratt @gurunrao @jansupol @jgallimore @kazumura @kwsutter @LanceAndersen @bhatpmk @RohitKumarJain @shighbar @gthoman @brideck @OndroMih @dmatej
@starksm64 @scottmarlow

PR #2396, opened 2025-07-27
Defend against intermittent EJB failures

dmatej:
**Fixes Issue**
See https://github.com/jakartaee/platform-tck/pull/2322 - this is a backport to 10.0.x (7)

```
Finished Test:  FAILED........com/sun/ts/tests/ejb32/lite/timer/interceptor/lifecycle/singleton/JsfClient.java#postConstructInBeanClass_from_ejblitejsf
```

@starksm64 @scottmarlow

### DIFF
--- a/src/com/sun/ts/tests/ejb30/common/lite/EJBLiteClientBase.java
+++ b/src/com/sun/ts/tests/ejb30/common/lite/EJBLiteClientBase.java
@@ -22,6 +22,8 @@ package com.sun.ts.tests.ejb30.common.lite;
 import java.io.File;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.util.ArrayList;
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Properties;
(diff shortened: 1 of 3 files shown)


dmatej:
> @dmatej this change should only be made for EE 12 (e.g. main) as per https://jakarta.ee/committees/specification/tckprocess We do have a TCK challenge process that could be used to consider excluding or updating invalid tests.
> 
> Tests that fail intermittently are annoying but probably not considered invalid.
> 
> Update: since it was already made for EE 11, it is fixed for EE 12 already as well.

Maybe I am missing something, but this change doesn't exclude tests, just fixes failure on Platform TCK 10 executed on JDK 17. I am now creating merge of GF master to GF 8.0, it would be good if all TCK branches/versions would give reliable results.

dmatej:
Aha, so I am doing something wrong, I see. I assumed that simple bugs could be just reproduced, fixed and the fix retested, mea culpa.

> How Tests May be Added or Changed in a TCK
> Tests may be added, updated, removed or excluded as the specification team sees fit in major or minor updates.
> 
> In service releases, specification teams should only make changes that are documented in TCK challenge issues, and service release notes should document which challenges have been addressed. Test updates within a service release should follow the guidance described under Accepted Challenges.
> 

It would be good if PRs then would always contain links to TCK challenge issues then. #2322 doesn't, so I thought I can just simply backport it.

I will revert it, however it is a bug in TCK 10, so sure, I am interested in how to get the fix here. The last release is 10.0.6, what is the process to release 10.0.7 (with this fix included)?

PR #2654, opened 2026-03-27
[2653] Do not use the java.version system property when attempting to…

### LINKED ISSUE #2653: The Tags TCK checks the `java.version` property assuming there will always be a . in the version.
**Specification**: Jakarta Tags 

**Tests**:
* `com.sun.ts.tests.jstl.compat.onedotzero.JSTLClientIT`
* `com.sun.ts.tests.jstl.spec.fmt.format.localecontext.JSTLClientIT`
* `com.sun.ts.tests.jstl.spec.fmt.format.parsedate.JSTLClientIT`
* `com.sun.ts.tests.jstl.spec.fmt.format.settimezone.JSTLClientIT`
* `com.sun.ts.tests.jstl.spec.fmt.format.timezone.JSTLClientIT`
* `com.sun.ts.tests.jstl.spec.fmt.i18n.bundle.JSTLClientIT`

**Implementation**:
WildFly

**Issue**:
The tests themselves are okay, but there is a utility in the `com.sun.ts.tests.jstl.common.client.AbstractUrlClient` for checking if the Java version is 20 or higher. The problem is this uses the `java.version` system property and expects it always contains a `.` which is not always the case.

Bad Method:
```java
public boolean isJavaVersion20OrGreater() {
  boolean isJavaVersion20OrGreater = false;

  String version = System.getProperty("java.version");
  int majorVersionDot = version.indexOf(".");

  version = version.substring(0, majorVersionDot);

  if (Integer.parseInt(version) >= 20) {
      isJavaVersion20OrGreater = true;
  }

  return isJavaVersion20OrGreater;
}
```

Suggested Change:
```java
public boolean isJavaVersion20OrGreater() {
  return Runtime.version().feature() >= 20;
}
```

This is a minor change and is portable.

jamezp:
… determine the Java version. Instead use the Runtime.version() in the Jakarta Tags TCK.

**Fixes Issue**
fixes #2653 

**Describe the change**
Uses the `Runtime.vesion()` to determine the JVM version rather than the `java.version` system property.

### DIFF
--- a/tcks/apis/tags/src/main/java/com/sun/ts/tests/jstl/common/client/AbstractUrlClient.java
+++ b/tcks/apis/tags/src/main/java/com/sun/ts/tests/jstl/common/client/AbstractUrlClient.java
@@ -229,18 +229,7 @@ public class AbstractUrlClient extends BaseUrlClient {
   }
 
   public boolean isJavaVersion20OrGreater() {
-    boolean isJavaVersion20OrGreater = false;
-
-    String version = System.getProperty("java.version");
-    int majorVersionDot = version.indexOf(".");
-
-    version = version.substring(0, majorVersionDot);
-
-    if (Integer.parseInt(version) >= 20) {
-        isJavaVersion20OrGreater = true;
-    }
-
-    return isJavaVersion20OrGreater;
+    return Runtime.version().feature() >= 20;
   }
 
 }


pnicolucci:
I'm ok with this change, see the following comment for details: https://github.com/jakartaee/platform-tck/issues/2653#issuecomment-4169408670

pnicolucci:
@scottmarlow, for EE12, we're removing the tags TCK from the platform-tck, right? Should this instead be a PR against: https://github.com/jakartaee/tags? Do we need a PR to remove the tags tck from the platform-tck for EE12? I know there was some discussion here: https://github.com/jakartaee/tags/issues/270

scottmarlow:
> @scottmarlow, for EE12, we're removing the tags TCK from the platform-tck, right? 

Oh right, thanks for the reminder!  We already moved Tags tests over to https://github.com/jakartaee/tags/tree/main/tck/src/main/java/com/sun/ts/tests/jstl so :+1: 

> Should this instead be a PR against: https://github.com/jakartaee/tags? 

Yes! 

> Do we need a PR to remove the tags tck from the platform-tck for EE12? 

Yes

> I know there was some discussion here: [jakartaee/tags#270](https://github.com/jakartaee/tags/issues/270)

:+1:

scottmarlow:
Merging even though the Tags TCK source has been moved to the Tags spec repo.

jamezp:
FYI I created the upstream PR https://github.com/jakartaee/tags/pull/287.

PR #2682, opened 2026-04-08
Fix mixed access strategy in Department4 entities (Issue #2678 , main branch) 

### LINKED ISSUE #2678: TCK Department4 uses mixed access strategy without @Access, leading to undefined behavior
**Describe the issue**  
The `Department4` entity used by the Jakarta Persistence TCK `mapkeytemporal` test mixes field and property annotation placement without explicitly declaring a class-level access type via `@Access`.

In the current source:
- `@Id` is placed on the field `id`
- `@Transient` is placed on the field `lastNameEmployees`
- mapping annotations for `lastNameEmployees` are placed on the getter method
- no class-level `@Access` annotation is declared

According to Jakarta Persistence 3.2, default access is inferred from annotation placement (https://jakarta.ee/specifications/persistence/3.2/jakarta-persistence-spec-3.2#default-access-type), and classes whose access type is defaulted must be consistent in their placement of mapping annotations. The spec further states that behavior is undefined when annotation placement is mixed on fields and properties within an entity hierarchy without explicitly specifying the class-level `Access` annotation.

As a result, this TCK test appears to depend on undefined behavior under Jakarta Persistence 3.2.

---

**To Reproduce**  
1. Open  
   ~~`tcks/apis/persistence/persistence-inside-container/spec-tests/src/main/java/ee/jakarta/tck/persistence/core/annotations/mapkeytemporal/Department4.java`~~

`tcks/apis/persistence/persistence-inside-container/spec-tests/src/main/java/ee/jakarta/tck/persistence/core/annotations/mapkeyenumerated/Department4.java`

2. Observe:
   - `@Id` on field `id`
   - `@Transient` on field `lastNameEmployees`
   - `@ElementCollection` and related mapping annotations on `getLastNameEmployees()`
3. Observe that no class-level `@Access` is declared

---

**Expected behavior**  
The entity should either:
- use a consistent default access strategy, or
- explicitly declare mixed access using class-level `@Access(...)` with attribute-level override where needed

TCK tests should rely only on behavior defined by the Jakarta Persistence specification.

---

**Relevant specification text**  
Jakarta Persistence 3.2, section 2.3.1 states that:
- default access is inferred from annotation placement
- classes whose access type is defaulted must be consistent in annotation placement
- behavior is undefined when annotations are mixed on fields and properties without explicitly specifying the class-level `Access` annotation

Section 2.3.2 defines attribute-level access override for a class that explicitly specifies an access type using `@Access`.

---

**Potential improvement**  
Make the intended mixed access explicit, for example:

```java
@Entity
@Access(AccessType.FIELD)
@Table(name = "DEPARTMENT2")
public class Department4 {

    @Id
    private int id;

    private String name;

    @Transient
    private Map<Date, EmbeddedEmployee> lastNameEmployees;

    @Access(AccessType.PROPERTY)
    @ElementCollection(targetClass = EmbeddedEmployee.class)
    @CollectionTable(name = "EMP_MAPKEYCOL2", joinColumns = @JoinColumn(name = "FK_DEPT5"))
    @AttributeOverrides({
        @AttributeOverride(name = "employeeId", column = @Column(name = "ID")),
        @AttributeOverride(name = "employeeName", column = @Column(name = "LASTNAME"))
    })
    @MapKeyEnumerated(EnumType.STRING)
    public Map<Date, EmbeddedEmployee> getLastNameEmployees() {
        return lastNameEmployees;
    }
}

### PR DESCRIPTION
**Pull request by kuuku123**
**Fixes Issue**
Resolves https://github.com/jakartaee/platform-tck/issues/2678

**Describe the change**
the entity has been simplified to use a consistent, implicit FIELD access strategy.

@starksm64 @scottmarlow

### DIFF
--- a/tcks/apis/persistence/persistence-inside-container/spec-tests/src/main/java/ee/jakarta/tck/persistence/core/annotations/mapkeyenumerated/Department4.java
+++ b/tcks/apis/persistence/persistence-inside-container/spec-tests/src/main/java/ee/jakarta/tck/persistence/core/annotations/mapkeyenumerated/Department4.java
@@ -29,10 +29,12 @@ import jakarta.persistence.Id;
 import jakarta.persistence.JoinColumn;
 import jakarta.persistence.MapKeyEnumerated;
 import jakarta.persistence.Table;
-import jakarta.persistence.Transient;
+import jakarta.persistence.Access;
+import jakarta.persistence.AccessType;
 
 @Entity
 @Table(name = "DEPARTMENT2")
+@Access(AccessType.FIELD)
 public class Department4 implements java.io.Serializable {
 
 	private static final long serialVersionUID = 22L;
@@ -43,7 +45,11 @@ public class Department4 implements java.io.Serializable {
 
 	private String name;
 
-	@Transient
+	@ElementCollection(targetClass = EmbeddedEmployee.class)
+	@CollectionTable(name = "EMP_MAPKEYCOL2", joinColumns = @JoinColumn(name = "FK_DEPT5"))
+	@AttributeOverrides({ @AttributeOverride(name = "employeeId", column = @Column(name = "ID")),
+			@AttributeOverride(name = "employeeName", column = @Column(name = "LASTNAME")) })
+	@MapKeyEnumerated(EnumType.STRING)
 	private Map<Numbers, EmbeddedEmployee> lastNameEmployees;
 
 	public Department4() {
@@ -78,11 +84,6 @@ public class Department4 implements java.io.Serializable {
 	// ===========================================================
 	// getters and setters for the association fields
 
-	@ElementCollection(targetClass = EmbeddedEmployee.class)
-	@CollectionTable(name = "EMP_MAPKEYCOL2", joinColumns = @JoinColumn(name = "FK_DEPT5"))
-	@AttributeOverrides({ @AttributeOverride(name = "employeeId", column = @Column(name = "ID")),
-			@AttributeOverride(name = "employeeName", column = @Column(name = "LASTNAME")) })
-	@MapKeyEnumerated(EnumType.STRING)
 	public Map<Numbers, EmbeddedEmployee> getLastNameEmployees() {
 		return lastNameEmployees;
 	}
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: tcks/apis/persistence/persistence-inside-container/spec-tests/src/main/java/ee/jakarta/tck/persistence/core/annotations/mapkeyenumerated/Department4.java [hunk 1]
```diff
@@ -43,7 +45,11 @@ public class Department4 implements java.io.Serializable {
 
 	private String name;
 
-	@Transient
```
scottmarlow: Please keep the `@Transient` on this field.

kuuku123:
following is overall summary of this PR.
 
1. **`Department4` had undefined access semantics.**  
   The entity mixed field and property annotation placement without an explicit `@Access`, which is undefined under Jakarta Persistence 3.2. This was normalized to field access so the mapping is explicit and spec-compliant.

2. **The map key column needed to be made explicit.**  
   Once `lastNameEmployees` became an effective field-mapped `@ElementCollection`, the provider expected a map-key column. Without an explicit `@MapKeyColumn`, it fell back to the default column name (`[ERROR: column "lastnameemployees_key"` ), 
   but the Derby DDL did not define that column. This was fixed by adding `@MapKeyColumn` and updating the DDL accordingly.

3. **The embeddable value overrides were corrected to match the spec.**  
   `@AttributeOverrides` for the map value embeddable were updated to use the `value.` prefix, so the override paths align with the specification for map value attributes.

4. **The test setup was adjusted to avoid a second, unrelated table-sharing conflict.**  
   After the field-access fix, `Department4.lastNameEmployees` became a real `@ElementCollection` mapping to `EMP_MAPKEYCOL2`. But `Employee4` was also being persisted as an entity to the same physical table, causing duplicate inserts on the same primary key. Since `Employee4` was only needed as a query entry point in `elementCollectionTest`, that test was changed to load `Department4` directly and the separate `Employee4` persistence path was removed.
   
   
   
and it did pass `ee.jakarta.tck.persistence.core.annotations.mapkeyenumerated.ClientPmservletTest`
 ```
Apr 14, 2026 9:39:33 AM tck.arquillian.protocol.appclient.DeploymentMonitor undeploy
INFO: jpa_core_annotations_mapkeyenumerated_pmservlet_vehicle, undeployed
[INFO] Tests run: 4, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 9.525 s -- in ee.jakarta.tck.persistence.core.annotations.mapkeyenumerated.ClientPmservletTest
Stopping jeus.ctsport.arquillian.JEUSManagedDeployableContainer@5f780a86...
Stopping jeus.ctsport.arquillian.JEUSManagedDeployableContainer@34819867...
Stopping jeus.ctsport.arquillian.JEUSManagedDeployableContainer@118102ee...
[INFO]
[INFO] Results:
[INFO]
[INFO] Tests run: 4, Failures: 0, Errors: 0, Skipped: 0
```

PR #2698, opened 2026-04-24
[2667] Follows up by changing the URI's for the tag libraries to the …

jamezp:
…correct URI for the web-app version. This change was made in https://github.com/jakartaee/tags/issues/144 and the old web-app URI enabled the old namespace.

**Fixes Issue**
This is a follow up to #2667 
resolves #2667 

**Related Issue(s)**
Related to https://github.com/jakartaee/tags/issues/144

**Describe the change**
This PR removes the duplicate `jstl_1_0_compat_web.xml` file. It also changes the `uri` attributes for the `talib` to the new URI. When we upgrade the `web-app` namespace version, this triggered the requirement of the new taglib namespace. 

It also has a small change to remove the `<display-name/>` from the remaining `jstl_1_0_compat_web.xml` as that element is no longer supported. This was already done for Jakarta EE 10 in https://github.com/jakartaee/platform-tck/pull/1335.

I will note this package name is specifically named `com.sun.ts.tests.jstl.compat.onedotzero`. However, I see nothing in the JavaDocs of the test, `com.sun.ts.tests.jstl.compat.onedotzero.JSTLClientIT`, that seems to indicate it's actually testing some kind of backwards compatibility.

That said, if I'm wrong we need to revert:

- https://github.com/jakartaee/platform-tck/pull/2681

I have tested this change on WildFly and GlassFish and they both pass. They both failed with the current 11.0.3-SNAPSHOT, e.g. without this fix.

PR #2699, opened 2026-04-24
[2667] Follows up by changing the URI's for the tag libraries to the …

### LINKED ISSUE #2667: TCK: tags-tck depends on DTD-based deployment descriptors
**Challenges Tests**
  - com.sun.ts.tests.jstl.compat.onedotzero.JSTLClientIT

**TCK Version**
Jakarta EE TCK 11.0.1

**Description**
The com.sun.ts.tests.jstl.compat.onedotzero.JSTLClientIT  test class tries to deploy an applicatation with an old web.xml that contains a DTD reference.  We already have an accepted TCK challenge for this in Servlet (https://github.com/jakartaee/servlet/issues/1004).

jamezp:
…correct URI for the web-app version. This change was made in https://github.com/jakartaee/tags/issues/144 and the old web-app URI enabled the old namespace.

Upstream PR #2698 

**Fixes Issue**
This is a follow up to #2667 
resolves #2667 

**Related Issue(s)**
Related to https://github.com/jakartaee/tags/issues/144

**Describe the change**
This PR removes the duplicate `jstl_1_0_compat_web.xml` file. It also changes the `uri` attributes for the `talib` to the new URI. When we upgrade the `web-app` namespace version, this triggered the requirement of the new taglib namespace. 

It also has a small change to remove the `<display-name/>` from the remaining `jstl_1_0_compat_web.xml` as that element is no longer supported. This was already done for Jakarta EE 10 in https://github.com/jakartaee/platform-tck/pull/1335.

I will note this package name is specifically named `com.sun.ts.tests.jstl.compat.onedotzero`. However, I see nothing in the JavaDocs of the test, `com.sun.ts.tests.jstl.compat.onedotzero.JSTLClientIT`, that seems to indicate it's actually testing some kind of backwards compatibility.

That said, if I'm wrong we need to revert:

- https://github.com/jakartaee/platform-tck/pull/2681
- https://github.com/jakartaee/platform-tck/pull/2684

I have tested this change on WildFly and GlassFish and they both pass. They both failed with the current 11.0.3-SNAPSHOT, e.g. without this fix.


scottmarlow:
I also verified this change locally as well.  Thanks @jamezp !

CC @pnicolucci

pnicolucci:
@volosied  and I will review this PR on Monday and let you know if we have any questions or concerns. Thanks!

volosied:
Hi, I don't have any objects to this change, but I'm curious how the two servers failed on 11.0.3-SNAPSHOT? 

The [META-INF contains the TLDs for both namespaces](https://github.com/jakartaee/tags/tree/3.0.1-IMPL-RELEASE/impl/src/main/resources/META-INF). Shouldn't the JSP container pick up  the all the TLDs (i.e both namespaces)? 

We let both URIs exist as a transition, but will work to remove the older URIs in EE12.

jamezp:
I'll have a look at the implementation to see how it discovers those. The error we were seeing without this change were around expressions in the tag attributes, for example:
```
<!-- EL - Simple validation that when's body content is processed
     if the test evaluates to true -->
<c:choose>
    <c:when test="${elTrue}">
        PASSED<br>
    </c:when>
</c:choose>
```

The expression in the "test" attribute seemed to be an issue. However, using an older `web-app` version or updating the `uri` attribute for the tag import seems to fix it.

I will hopefully have some time this afternoon to look to see how it works.

scottmarlow:
@volosied

> Hi, I don't have any objects to this change, but I'm curious how the two servers failed on 11.0.3-SNAPSHOT?
> 
> The [META-INF contains the TLDs for both namespaces](https://github.com/jakartaee/tags/tree/3.0.1-IMPL-RELEASE/impl/src/main/resources/META-INF). Shouldn't the JSP container pick up the all the TLDs (i.e both namespaces)?
> 
> We let both URIs exist as a transition, but will work to remove the older URIs in EE12.

I'm not sure but it looks like https://github.com/jakartaee/tags/commit/3687591dcc3075af7f39752b558fd4fed30837a4 does seem to include the old Sun TLDs but does the Tags specification require that the old Sun TLDs/URIs work with deployments using Jakarta EE [web-app_5_0.xsd](https://jakarta.ee/xml/ns/jakartaee/web-app_5_0.xsd) or newer?  It seems that Jakarta EE 11 implementations have already passed the Jakarta EE 11 Platform TCK with the old Sun based web-app dtd but our TCK change to instead update those tests to use a Jakarta EE web-app schema break those already passing implementations.  Thankfully, James tested the other implementation.  We also know that WildFly also fails as well which James is going to look into but I'm questioning now how to move forward with the EE 11 Platform TCK 11.0.3 release and these Tags compat tests.

For reference, the compat TCK tests have these assertions:

```
<assertion defined-by="technology" impl-spec="false" 
                   required="true" status="active" testable="true"> 
                   <id>JSTL:spec:106.3</id> 
                   <description>Otherwise, the i18n localization context whose 
                       resource bundle is to be used for localization is specified via 
                       the javax.servlet.jsp.jstl.fmt.localizationContext configuration 
                       setting.</description> 
                   <location chapter="8" section="8.2" /> 
                   <sub-assertions> 
                       <assertion defined-by="technology" impl-spec="false" 
                           required="true" status="active" testable="true"> 
                           <id>JSTL:spec:106.3.1</id> 
                           <description>If the configuration setting is of type 
                               LocalizationContext its resource bundle component is used for 
                               localization.</description> 
                           <location chapter="8" section="8.2" /> 
                       </assertion> 
                       <assertion defined-by="technology" impl-spec="false" 
                           required="true" status="active" testable="true"> 
                           <id>JSTL:spec:106.3.2</id> 
                           <description>Otherwise, the configuration setting is of type 
                               String, and the action establishes its own i18n localization 
                               context whose resource bundle component is determined according 
                               to the resource bundle determination algorithm in Section 8.3, 
                               using the configuration setting as the resource bundle basename.</description> 
                           <location chapter="8" section="8.2" /> 
                       </assertion> 
                   </sub-assertions> 
               </assertion>
```

I look forward to understanding more about the requirements but we will need to exclude these `compat` tests I think soon or merge this pull request so that the already passing GlassFish implementation continues to work.

Again, thanks James for also testing locally with GlassFish which passes with this pr!

CC @arjantijms

scottmarlow:
This has been open for a while and its not exactly clear of how updated com.sun.ts.tests.jstl.compat.onedotzero tests would validate current Jakarta EE 11 Platform requirements, I'm thinking we should exclude the compat.onedotzero tests in the next few days.

jamezp:
This next week I should have time to look into what the implementation is doing and seeing if something odd happens there. 

I do know these tests fail on both WildFly and Glassfish without this specific change.

jamezp:
I don't really know yet why changing the `web-app` namespace causes the issue. However, I did notice that https://github.com/jakartaee/tags/blob/c48a7a0966768908fc4a34a7260682c3edc373c8/impl/src/main/resources/META-INF/c-1_0.tld#L133-L137 does have a ` <rtexprvalue>false</rtexprvalue>` and the error I see is:
```
JBWEB004195: According to TLD or attribute directive in tag file, attribute test does not accept any expressions
```

jamezp:
It looks like we do need this change or we need to revert back to the old namespace. What I see happening is when the web-app version is less than 2.4 it defaults the `defaultIsELIgnored` to `true` so the expressions are ignored when validating. However, when a newer version is used, expressions are validated and the older TLD file doesn't support expressions in the attribute. 

The test itself is in a package that ends with `compat.onedotzero` which makes me think it exists to test backwards compatibility. If that is the case, I do feel we need to revert the namespace back. Otherwise, we need this fix.

volosied:
Let's merge it in.

scottmarlow (approved):

PR #2703, opened 2026-05-12
Migrate CDI EE TCK assertions to CDI TCK 5.0.0.Alpha5 API

manovotn:
There were some changes in the CDI TCK test utils and as it turns out, those are used here as well.
Original TCK PR was https://github.com/jakartaee/cdi-tck/pull/676

These changes adapt tests under platform TCKs accordingly.
There is also a change in the TCK coordinates - all CDI and CDI TCK artifacts changed there coordinates to more sensible ones as part of the 5.0 development.

CC @Ladicek  @Azquelt

### DIFF
--- a/tcks/apis/cdi-ee-tck/tck/pom.xml
+++ b/tcks/apis/cdi-ee-tck/tck/pom.xml
@@ -38,9 +38,9 @@
     <dependencies>
         <!-- Jakarta EE dependencies -->
         <dependency>
-            <groupId>jakarta.enterprise</groupId>
-            <artifactId>cdi-tck-core-impl</artifactId>
-            <version>4.1.0</version>
+            <groupId>jakarta.cdi</groupId>
+            <artifactId>jakarta.cdi-tck-core-impl</artifactId>
+            <version>5.0.0.Alpha5</version>
         </dependency>
 
         <dependency>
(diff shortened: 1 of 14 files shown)


Ladicek (approved):

Azquelt (approved):

scottmarlow (approved):

scottmarlow:
Thanks @manovotn + @Ladicek  + @Azquelt!

arjantijms:
Just an observation; the cdi tests in the platform are always maintained and updated by members from the CDI team.

Isn't that an indication they actually belong in the CDI repo (in an EE folder, separate from the SE tests)?

manovotn:
> Just an observation; the cdi tests in the platform are always maintained and updated by members from the CDI team.
> 
> Isn't that an indication they actually belong in the CDI repo (in an EE folder, separate from the SE tests)?

These were deliberately removed from CDI repo in the first place so I am pretty sure we *do not want them back* :-)
I don't exactly recall all the reasons (@starksm64 did that I think?) but circular deps might have been one of them. 
These tests require dependencies on other specifications which is then problematic when it comes to releases as CDI is part of one of the first waves.

The reason other is simply the fact that these tests deal with integration of other EE specs with CDI. Ideally, these tests would be adopted by their respective specifications if/where applicable.

arjantijms:
> These tests require dependencies on other specifications which is then problematic when it comes to releases as CDI is part of one of the first waves.

If they are in a separate folder within the CDI GitHub repo, they don't have to be released.

arjantijms:
> Ideally, these tests would be adopted by their respective specifications if/where applicable.

Depending on what these tests are doing, if they are indeed mostly about say Jakarta Faces with CDI, then they should move there.

Do you have some candidates already in mind for moving?

manovotn:
> > Ideally, these tests would be adopted by their respective specifications if/where applicable.
> 
> Depending on what these tests are doing, if they are indeed mostly about say Jakarta Faces with CDI, then they should move there.
> 
> Do you have some candidates already in mind for moving?

Not from the top of my head, they'd have to be inspected individually. Or perhaps looking for some of the issues related to that move would give us a hint?

Though unfortunately it will be hard to find a ~victim~ volunteer to spend his/her cycles on going through identifying those tests and then and trying to get other specs to adopt them. You know how it goes - PRs are welcome :)
It might be easier with some coding agent perhaps, I didn't give that a go...

